PR #1, opened 2019-07-09
Make azure work maybe

rminnich:


azure-pipelines.yml
```diff
@@ -28,11 +28,8 @@ steps:
 
 - script: |
     go version
-    go get -v -t -d ./...
-    if [ -f Gopkg.toml ]; then
-        curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-        dep ensure
-    fi
+    git clone git@github.com:u-root/webboot
```
rjoleary: This looks like it's using the latest code from head rather than the code in the PR.

rminnich: that's a problem. How do I fix that? and why did it work, hmm. ah because this head and that head are same, only azure gits changed. OK see next repo.

rminnich:
ah this did it:
Note: checking out 'refs/remotes/pull/1/merge'.

rminnich:
we're not going to use modules for now.

rminnich:
ok gan options for build are in there now.
you can add u-root options, and also select the commands you want.

webboot.go
```diff
@@ -8,12 +8,10 @@ import (
...
-	debug    = func(string, ...interface{}) {}
-	verbose  = flag.Bool("v", true, "verbose debugging output")
-	commands = [][]string{
-		{"date"},
-		{"go", "run", "github.com/u-root/u-root/."},
-	}
+	debug   = func(string, ...interface{}) {}
+	verbose = flag.Bool("v", true, "verbose debugging output")
+	uroot   = flag.String("u", "", "options for u-root")
+	cmds    = flag.String("c", "github.com/u-root/u-root/cmds/core/*", "u-root commands to build into the image")
```
GanShun: What about using the core template rather than cmds/core/* I'm fine either way

GanShun: Also once Louis' stuff gets merged, we should also auto include webboot's commands.

rminnich: yeah, there's a problem doing that but I forget what it is, let me try it.

rminnich: hmm seems to work now ... ok done.

rminnich: next time out we'll add github.com/u-root/weboots/cmds/*

webboot.go
```diff
@@ -24,6 +22,13 @@ func init() {
 }
 
 func main() {
+	var commands = [][]string{
+		{"date"},
+		{"go", "get", "-u", "github.com/u-root/u-root"},
+		// Due to u-root bugs, we have to use bb mode and we have to use
```
GanShun: Remove comment, sorry I missed this

rminnich:
anyone?

webboot.go
rminnich: oops. fixed.

PR #2, opened 2019-07-22
Add the weboot.go file where the web boot code will go.

louismurerwa:
Added the webboot command to u-root so the command appears in u-root.


loadWebboot.go
```diff
@@ -12,7 +12,7 @@ var (
 	verbose  = flag.Bool("v", true, "verbose debugging output")
```
rminnich: I would prefer a name like
buildimage.go

louismurerwa: Done

PR #3, opened 2019-07-24
Webboot's Dhclient

upatel99:
Implement the dhclient command to accommodate webboot.

### DIFF
--- a/integration/dhclient_test.go
+++ b/integration/dhclient_test.go
@@ -1,4 +1,4 @@
-// Copyright 2018 the u-root Authors. All rights reserved
+// Copyright 2019 the u-root Authors. All rights reserved
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
@@ -51,11 +51,11 @@ func TestDhclient(t *testing.T) {
 		Cmds: []string{
 			"github.com/u-root/u-root/cmds/core/ip",
 			"github.com/u-root/u-root/cmds/core/init",
-			"github.com/u-root/u-root/cmds/core/dhclient",
+			"github.com/u-root/webboot/integration/testcmd/dhclient",
 			"github.com/u-root/u-root/cmds/core/shutdown",
 		},
 		Uinit: []string{
-			"dhclient -ipv6=false -v",
+			"dhclient",
 			"ip a",
 			"sleep 5",
 			"shutdown -h",
--- a/integration/testcmd/dhclient/main.go
+++ b/integration/testcmd/dhclient/main.go
@@ -2,23 +2,17 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Get the time the machine has been up
-// Synopsis:
-//     webboot
 package main
 
 import (
 	"flag"
-	"log"
 
 	"github.com/u-root/webboot/pkg/dhclient"
 )
 
 var (
-	ifName  = flag.String("interface", "^e.*", "Name of the interface")
 	timeout = flag.Int("timeout", 15, "Lease timeout in seconds")
 	retry   = flag.Int("retry", 5, "Max number of attempts for DHCP clients to send requests. -1 means infinity")
-	verbose = flag.Bool("verbose", false, "Verbose output")
 	ipv4    = flag.Bool("ipv4", true, "use IPV4")
 	ipv6    = flag.Bool("ipv6", true, "use IPV6")
 )
@@ -26,6 +20,5 @@ var (
 func main() {
 	flag.Parse()
 
-	dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
-	log.Println("weboot up")
+	dhclient.Request("^e.*", *timeout, *retry, true, *ipv4, *ipv6)
 }
--- a/pkg/dhclient/dhclient.go
+++ b/pkg/dhclient/dhclient.go
@@ -1,0 +1,81 @@
+// Copyright 2013-2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package dhclient
+
+import (
+	"context"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/u-root/u-root/pkg/dhclient"
+	"github.com/vishvananda/netlink"
+)
+
+// Request sets up the dhcp configurations for all of the ifNames.
+func Request(ifName string, timeout int, retry int, verbose bool, ipv4 bool, ipv6 bool) {
+	ifRE := regexp.MustCompilePOSIX(ifName)
+
+	ifnames, err := netlink.LinkList()
+	if err != nil {
+		log.Fatalf("Can't get list of link names: %v", err)
+	}
+
+	var filteredIfs []netlink.Link
+	for _, iface := range ifnames {
+		if ifRE.MatchString(iface.Attrs().Name) {
+			filteredIfs = append(filteredIfs, iface)
+		}
+	}
+
+	if len(filteredIfs) == 0 {
+		log.Fatalf("No interfaces match %s", ifName)
+	}
+
+	cl := make(chan error)
+	go configureAll(filteredIfs, cl, timeout, retry, verbose, ipv4, ipv6)
+	result := <-cl
+	log.Printf("Configuring DHCP returns with error: %v", result)
+}
+
+func configureAll(ifs []netlink.Link, cl chan<- error, timeout int, retry int, verbose bool, ipv4 bool, ipv6 bool) {
+	packetTimeout := time.Duration(timeout) * time.Second
+
+	ctx, cancel := context.WithTimeout(context.Background(), packetTimeout*time.Duration(1<<uint(retry)))
+	defer cancel()
+
+	c := dhclient.Config{
+		Timeout: packetTimeout,
+		Retries: retry,
+	}
+	if verbose {
+		c.LogLevel = dhclient.LogSummary
+	}
+	r := dhclient.SendRequests(ctx, ifs, ipv4, ipv6, c)
+
+	defer close(cl)
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Printf("Done with dhclient: %v", ctx.Err())
+			return
+
+		case result, ok := <-r:
+			if !ok {
+				log.Printf("Configured all interfaces.")
+				return
+			}
+			if result.Err != nil {
+				log.Printf("Could not configure %s: %v", result.Interface.Attrs().Name, result.Err)
+			} else if err := result.Lease.Configure(); err != nil {
+				log.Printf("Could not configure %s: %v", result.Interface.Attrs().Name, err)
+			} else {
+				cl <- nil
+				log.Printf("Configured %s with %s", result.Interface.Attrs().Name, result.Lease)
+			}
+		}
+	}
+}


webboot/webboot.go
```diff
@@ -0,0 +1,92 @@
...
+	if len(filteredIfs) == 0 {
+		log.Fatalf("No interfaces match %s", ifName)
+	}
+
+	channel := make(chan bool)
+	go configureAlls(filteredIfs,channel)
+	x := <-channel
+}
+
+func configureAlls(ifs []netlink.Link, channel chan bool) {
```
rminnich: make this a chan of errors, then you can push nil errors to say "it worked" and errors are useful

webboot/webboot.go
```diff
@@ -8,9 +8,26 @@
 package main
 
 import (
-	"fmt"
+	"flag"
+	"log"
+
+	"github.com/u-root/webboot/pkg/webbootDhclient"
```
rminnich: rename it to dhclient

pkg/webbootDhclient/webboot_dhclient.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/u-root/u-root/pkg/dhclient"
+	"github.com/vishvananda/netlink"
+)
+
+func DhclientNew(ifName string, timeout int, retry int, verbose bool, ipv4 bool, ipv6 bool) {
```
rminnich: just call this New. Just call the file dhclient.go. Just make the directory name dhclient. 
Make all file names all lower case.
So 
New()

upatel99: Resolved :)! Good point!

webboot/webboot.go
upatel99: makes sense!

webboot/webboot.go
upatel99: Good idea :)!

integration/testcmd/dhclient/main.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/u-root/webboot/pkg/dhclient"
+)
+
+func main() {
+	//call pkg/dhclient's New() with these parameters 
+	// ^e.* = interface name 
+	// 15 = timeOut in seconds
+	// 5 = number of retries 
+	// false = verbose
+	// the last true's are for searching both ipv4 and ipv6
```
rminnich: you can take this comment block out. If people need to know what it is they can use go doc.

pkg/dhclient/dhclient.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+	"log"
+	"regexp"
+	"time"
+
+	"github.com/u-root/u-root/pkg/dhclient"
+	"github.com/vishvananda/netlink"
+)
+
```
rminnich: you need a comment on this New function since it is exported. 
Be sure to run golint and go vet on these files before you push again.

webboot/webboot.go
```diff
@@ -8,9 +8,26 @@
...
+	ipv4    = flag.Bool("ipv4", true, "use IPV4")
+	ipv6    = flag.Bool("ipv6", true, "use IPV6")
 )
 
 func main() {
-	fmt.Println("weboot up")
+	//	flag.Parse()
+	//	if len(flag.Args()) > 1 {
+	//		log.Fatalf("Only one agument supported. Got %v, want name of an ISO", flag.Args())
+	//	}
```
rminnich: either remove this comment block or use it for something.

pkg/dhclient/dhclient.go
rminnich: Also, New by convention returns data structures or interfaces. This is not a New. This is actually doing the dhcp setup. It should have a different name.

integration/dhclient_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 the u-root Authors. All rights reserved
+// Copyright 2019 the Webboot Authors. All rights reserved
```
GanShun: I forget the ruling on this Ron, should this be changed?

integration/testcmd/dhclient/main.go
```diff
@@ -0,0 +1,14 @@
...
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"github.com/u-root/webboot/pkg/dhclient"
+)
+
+func main() {
+	dhclient.New("^e.*", 15, 5, false, true, true)
```
GanShun: I don't think you need to make the others flags, but I think it would be nice to make ipv4 and ipv6 flags, then default them to true.

GanShun: This way you can add tests for different configurations to test in the future.

pkg/dhclient/dhclient.go
GanShun: +1 on this. When I call something called dhclient.New, I expect that function to create a new instance of dhclient, not to do the whole protocol. You could call it something like Request or something.

integration/dhclient_test.go
rminnich: I have no clue but it might as well be lower case

pkg/dhclient/dhclient.go
```diff
@@ -0,0 +1,79 @@
...
+		log.Fatalf("No interfaces match %s", ifName)
+	}
+
+	channel := make(chan error)
+	go configureAlls(filteredIfs, channel, ifName, timeout, retry, verbose, ipv4, ipv6)
+	x := <-channel
+	log.Printf("The channel returns: %v", x)
+}
+
+func configureAlls(ifs []netlink.Link, channel chan error, ifName string, timeout int, retry int, verbose bool, ipv4 bool, ipv6 bool) {
```
rminnich: what is the purpose of this name ... configureAlls? What is the Alls here? All is already plural.

upatel99: this function is that looks for ipv4, ipv6, retries it 5 times, and times out with each retry 15 seconds added to it. It sends request to get connected. Once it finds and configures the interface with IPv4 or IPv6. It will print that. If it cannot find IPv4 or IPv6, then it will let us know that as well.

integration/dhclient_test.go
upatel99: I will lower case the Webboot

pkg/dhclient/dhclient.go
upatel99: Maybe, a good name for the internal function should remain to be configureAll without the s. However, let me try to come up with a better exported name :) !

pkg/dhclient/dhclient.go
upatel99: that is very true. I will try to come up with a better name for this function.

integration/testcmd/dhclient/main.go
upatel99: added some flags to this file. I made flags such as ipv4, ipv6, retry, and timeout to give people the options to try a few different things in the future. However, i left things like verbose set to true always and not as a flag as well as the interface name to "^e.*"

integration/dhclient_test.go
upatel99: actually, after our discussion, I changed the licensing rights back to u-root as this sounds the most applicable.

webboot/webboot.go
```diff
@@ -8,9 +8,24 @@
...
 import (
-	"fmt"
+	"flag"
+	"log"
+
+	"github.com/u-root/webboot/pkg/dhclient"
+)
+
+var (
+	ifName  = flag.String("InterfaceName", "^e.*", "Name of the interface")
```
rjoleary: There's an undocumented convention where flags are lowercase and short. Perhaps, name this flag "interface" rather than "InterfaceName".

pkg/dhclient/dhclient.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if len(filteredIfs) == 0 {
+		log.Fatalf("No interfaces match %s", ifName)
+	}
+
+	channel := make(chan error)
+	go configureAll(filteredIfs, channel, timeout, retry, verbose, ipv4, ipv6)
+	x := <-channel
+	log.Printf("The channel returns: %v", x)
```
rjoleary: Use a more informative error message. Maybe "Configuring DHCP failed with error: %v"?

pkg/dhclient/dhclient.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func configureAll(ifs []netlink.Link, channel chan error, timeout int, retry int, verbose bool, ipv4 bool, ipv6 bool) {
```
rjoleary: Give the channel parameter a better name, something like "results".

rjoleary: The type can also be the send-only channel type "chan<- error".

integration/testcmd/dhclient/main.go
```diff
@@ -0,0 +1,24 @@
...
+	"flag"
+
+	"github.com/u-root/webboot/pkg/dhclient"
+)
+
+var (
+	timeout = flag.Int("timeout", 15, "Lease timeout in seconds")
+	retry   = flag.Int("retry", 5, "Max number of attempts for DHCP clients to send requests. -1 means infinity")
+	ipv4    = flag.Bool("ipv4", true, "use IPV4")
+	ipv6    = flag.Bool("ipv6", true, "use IPV6")
```
rjoleary: You should either use these four flags or remove them.

pkg/dhclient/dhclient.go
```diff
@@ -0,0 +1,79 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			log.Printf("Done with dhclient: %v", ctx.Err())
+			return
+
+		case result, ok := <-r:
+			if !ok {
+				log.Printf("Configured all interfaces.")
+				return
```
rjoleary: Close the channel before returning. You can do a "defer close(channel)" near the start of this function to close it on all return paths.

integration/testcmd/dhclient/main.go
upatel99: I originally did not have them since I did not use any. However, Gan suggested that it would not hurt to add flags to the file main.go so you can add tests for different configurations to test in the future.

upatel99: I just had a discussion with you about this because I was slightly confused what you meant :) However, I get what you mean now! Thank you

pkg/dhclient/dhclient.go
upatel99: That is a good idea. As of right now, the channel could receive information to it. It is more type safety. With this, I can only send errors up. Thanks for the catch :D

pkg/dhclient/dhclient.go
upatel99: instead of "failed," I used "returns" because it does not always fail. Sometimes a nil error will get passed up. But, I made a more informative error message.

webboot/webboot.go
upatel99: Oh, sorry I did not realize that. Thank you for pointing it out. I will make the change.

pkg/dhclient/dhclient.go
upatel99: Oh, yes this is a good idea. As of right now, the channel is still open after having returned out of the function. I will make sure to close it. Thanks for the catch :)!

pkg/dhclient/dhclient.go
```diff
@@ -0,0 +1,81 @@
...
+		if ifRE.MatchString(iface.Attrs().Name) {
+			filteredIfs = append(filteredIfs, iface)
+		}
+	}
+
+	if len(filteredIfs) == 0 {
+		log.Fatalf("No interfaces match %s", ifName)
+	}
+
+	channel := make(chan error)
```
rminnich: just call it c not channel

upatel99: Changed it to cl as there is already c variable present in the file :)

webboot/webboot.go
upatel99: i removed the commented area. I believed I needed it to handle the "webboot [nameOfOS]" or "webboot [URL]" but then realized it was not necessary at the point for dhclient, so I commeted it out. But the comment is removed from the code as it is not necessary for dhclient. I will check for this flag length in the next PR

integration/testcmd/dhclient/main.go
upatel99: I have removed this comment out. I realize go doc will help describe this action by typing in go doc in the terminal and it shows the user the various different functions, variables, and flags.

webboot/webboot.go
upatel99: Sounds good :)!

pkg/webbootDhclient/webboot_dhclient.go
upatel99: I later changed the function name from New() to Request because one can expect this function to create a new instance of dhclient, not to do the whole protocol if it is called New()

PR #5, opened 2019-07-29
Add u-root integration's infrastructure to webboot

upatel99:
Louis and I copied the entire integration directory from u-root into
webboot. Then, since multiboot was currently failing, and we knew we
were not using it as of right now, we changed the "TestMultiBoot()" to
"testMultibootInternal()."

Update the README.md for better instructions to receive a kernel and
QEMU for the integration testing.

Signed-off-by: Urvisha Patel <urvishapatel@google.com>


integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Currently, only amd64 and arm are supported.
+
+## Steps to get the kernel (bzImage)
+#Step 1: clone the open source linux directory:
+git clone https://github.com/torvalds/linux.git
+
+#Step 2: enter the linux directory and see the directories and files within it
+cd linux
+ls
```
rminnich: you don't need to tell them to ls

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+example:
+    UROOT_TESTARCH=arm go test
+
+Currently, only amd64 and arm are supported.
+
+## Steps to get the kernel (bzImage)
+#Step 1: clone the open source linux directory:
+git clone https://github.com/torvalds/linux.git
+
+#Step 2: enter the linux directory and see the directories and files within it
```
rminnich: you don't need to say 'enter the directory'

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+When the QEMU arch is not amd64, set the `UROOT_TESTARCH` variable. For
+example:
+    UROOT_TESTARCH=arm go test
+
+Currently, only amd64 and arm are supported.
+
+## Steps to get the kernel (bzImage)
+#Step 1: clone the open source linux directory:
+git clone https://github.com/torvalds/linux.git
+
```
rminnich: just add a 'cd linux' command here

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+#Step 2: enter the linux directory and see the directories and files within it
+cd linux
+ls
+
+#Step 3: checkout the latest Linux version 
+git checkout v4.17
+
+#Step 4: preload a .config file. You need virtio_pci and virtio_net 
+wget https://raw.githubusercontent.com/u-root/u-root/master/.circleci/images/test-image-amd64/config_linux4.17_x86_64.txt
```
rminnich: No need to say 'You need' as that might change over time.
You can say
#Step 4: download a config file from u-root

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+#Step 5: build the kernel 
+Make -j12 
+
+#Step 6: visit your .bashrc and export the kernel and QEMU 
+Vi ~/.bashrc
+Export the following: 
+#NOTE: THIS IS JUST AN EXAMPLE! Make sure you put the right path to the kernel and QEMU by finding where they are located in your local machine.
+Example: 
+export UROOT_QEMU=”$HOME/bin/qemu-system-x86_64”
```
rminnich: You don't have to 'visit the rc and add this stuff'. That's up to the user. You can just say 
export UROOT_KERNEL="$HOME/linux/arch/x86_64/boot/bzImage"

And why do you assume they have a qemu in $HOME/bin? Almost nobody has that.

integration/RUNLOCAL
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# To use this script, drop a bzImage here and away you go.
+# Some examples:
+# ./RUNLOCAL
+# ./RUNLOCAL -v  -test.run TestTczclient
+UROOT_QEMU="qemu-system-x86_64" UROOT_KERNEL=bzImage go test "$@"
```
rminnich: see this usage? There's nothing set in .bashrc. It's set on the command line.

rminnich: And also note with this usage it's not pointing to $HOME/bin/qemu-system-x86_64

integration/README.md
rminnich: Don't they have to move this file to .config?

integration/README.md
```diff
@@ -0,0 +1,54 @@
+# Integration Tests
+
+This tests core use cases for u-root such as:
```
rminnich: u-root or webbot?

integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- retrieving and kexec'ing a Linux kernel,
+- uinit (user init), and
+- running unit tests requiring root privileges.
+
+## Usage
+
+Run the tests with:
+
+    go test
```
rminnich: or
UROOT_QEMU=path/to/qemu UROOT_KERNEL=path/to/kernel go test

integration/RUNLOCAL
GanShun: Think about why this works. Why does it work without the full location?

integration/README.md
upatel99: That is true

integration/README.md
upatel99: You're right. I had more details involved in the directions. I think I was more specific than needed to be.

integration/README.md
upatel99: sounds good!

integration/README.md
upatel99: You are right. We made the necessary changes.

integration/README.md
upatel99: Ours is in our /usr/, so we just made the changes by copying and pasting our export statements.

integration/README.md
upatel99: that works as well. However, using this command line, you would have to keep appending UROOT_QEMU and UROOT_KERNEL to go test every single time you want to go test using integration, correct?

integration/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+     or 
+
+    UROOT_QEMU=/usr/bin/qemu-system-x86_64 UROOT_KERNEL=$HOME/linux/arch/x86_64/boot/bzImage go test
```
rminnich: did you test to see if it needed the full path? You should do that.
Also indicate this is an example, i.e. put 
for example:
on the line above.

integration/README.md
```diff
@@ -0,0 +1,60 @@
...
+git checkout v4.17
+
+#Step 3: Download a config file from u-root and replace the current .config in linux with u-root .config file
+wget https://raw.githubusercontent.com/u-root/u-root/master/.circleci/images/test-image-amd64/config_linux4.17_x86_64.txt
+
+
+#Step 4: build the kernel 
+Make -j12 
+
+#Step 5: edit RUNLOCAL binary inside your integration folder to add a path to the kernel you built to as follows:
```
rminnich: no, they should not need to edit this file at all. All they need to do is set those two variables and run the file.

integration/README.md
rminnich: Further, you have this example command here. Why would you tell them to use RUNLOCAL if the command line you just showed them here will work?

integration/README.md
rminnich: in fact once they build the kernel they can use the command line from line 17, right?

integration/README.md
upatel99: Since there are multiple options to run integration testing, I provided a specific distinction in which one they can use

integration/README.md
upatel99: good point.

integration/README.md
upatel99: This should be webboot. We copied the interrogation package over from u-root for webboot's integration testing, but we did not catch the u-root there. We have changed the name since then.

integration/README.md
upatel99: In these distinction, the user was able to pick which way to use. However, I realize not everyone uses their bashrc and edit their bashrc to keep variable names in there.

integration/RUNLOCAL
upatel99: UROOT_QEMU uses the $PATH. Since qemu-system-x86_64 is stored inside of directory in $PATH, it knows exactly where to locate it. and this will work as long as you drop a bzImage (image of the kernel) inside because then it has the kernel and the qemu-system-x86_64 to run locally.

integration/README.md
upatel99: Ours is in our /usr/. We realize that no body would likely have it stored in their $HOME/bin. So we then made the changes and used the path where ours is stored (/usr/ ......) as an example for the readers.

integration/README.md
upatel99: Yes, since we do download the u-root's .config, we do have to move it to linux directory and then make the kernel using the new .config. We added that to the comments.

integration/README.md
upatel99: We realized one should not have to edit this to simply run it locally. Instead, yes, you are right, they can just use the command line from 17 to run locally.

upatel99: if they want to use the RUNLOCAL file, they can just drop a bzImage and just run "./RUNLOCAL" without editing this file, this can also work and it will run the integration test in that way as well.

PR #6, opened 2019-07-30
mountISO

louismurerwa:
Implemeted mountISO function in webboot package.
Adding Mount Tests

### DIFF
--- a/buildimage.go
+++ b/buildimage.go
@@ -1,3 +1,6 @@
+// Copyright 2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
 package main
 
 import (
--- a/integration/mountkexec_test.go
+++ b/integration/mountkexec_test.go
@@ -1,0 +1,112 @@
+// Copyright 2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+package integration
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/u-root/webboot/pkg/tlog"
+
+	"github.com/u-root/u-root/pkg/cp"
+	"github.com/u-root/u-root/pkg/golang"
+	"github.com/u-root/u-root/pkg/uroot"
+	"github.com/u-root/u-root/pkg/uroot/builder"
+	"github.com/u-root/u-root/pkg/uroot/initramfs"
+)
+
+// genISO generates an iso containing a kernel and initramfs to be used for testing mount and kexec
+func genISO(t *testing.T) (err error) {
+	// Copy kernel to tmpDir for tests involving kexec.
+	tmpDir, err := ioutil.TempDir("", "uroot-integration")
+	if err != nil {
+		return err
+	}
+	kernel := filepath.Join(tmpDir, "kernel")
+	if err := cp.Copy(os.Getenv("UROOT_KERNEL"), kernel); err != nil {
+		return err
+	}
+
+	// OutputFile
+	tlogger := &tlog.Testing{t}
+	newOutputFile := filepath.Join(tmpDir, "initiso.cpio")
+	w, err := initramfs.CPIO.OpenWriter(tlogger, newOutputFile, "", "")
+	if err != nil {
+		return err
+	}
+
+	cmds := []string{"github.com/u-root/u-root/cmds/*"}
+	env := golang.Default()
+	env.CgoEnabled = false
+	env.GOARCH = TestArch()
+
+	// Build u-root
+	opts := uroot.Opts{
+		Env: env,
+		Commands: []uroot.Commands{
+			{
+				Builder:  builder.BusyBox,
+				Packages: cmds,
+			},
+		},
+		TempDir:      tmpDir,
+		BaseArchive:  uroot.DefaultRamfs.Reader(),
+		OutputFile:   w,
+		DefaultShell: "elvish",
+	}
+
+	if err := uroot.CreateInitramfs(tlogger, opts); err != nil {
+		return err
+	}
+	o, err := exec.Command("genisoimage", "-o", "/tmp/tempdata.iso", tmpDir).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%v: %v", o, err)
+	}
+
+	return nil
+}
+
+// TestMountKexec runs an init which mounts a filesystem and kexecs a kernel.
+func TestKexecMount(t *testing.T) {
+	if err := genISO(t); err != nil {
+		t.Error(err)
+	}
+	if TestArch() != "amd64" {
+		t.Skipf("test not supported on %s", TestArch())
+	}
+
+	// Create the CPIO and start QEMU.
+	q, cleanup := QEMUTest(t, &Options{
+		Cmds: []string{
+			"github.com/u-root/webboot/integration/testcmd/mountkexec/uinit",
+			"github.com/u-root/u-root/cmds/core/init",
+			"github.com/u-root/u-root/cmds/core/ls",
+		}, Files: []string{
+			"/tmp/tempdata.iso:testIso",
+		},
+	})
+	defer cleanup()
+	// Loop through the expected results and compare with qemu's output
+	var results = []struct {
+		expect string
+	}{
+		{"Empty Directory Name"},
+		{"error making mount directory:mkdir : no such file or directory"},
+		{"Non-existent Iso"},
+		{"error setting loop device:open non-existentfile: no such file or directory"},
+		{"GoodTest"},
+		{"kernel"},
+	}
+
+	for _, results := range results {
+		if err := q.Expect(results.expect); err != nil {
+			t.Error(err)
+		}
+	}
+
+}
--- a/integration/testcmd/mountkexec/uinit/mountkexec.go
+++ b/integration/testcmd/mountkexec/uinit/mountkexec.go
@@ -1,0 +1,41 @@
+// Copyright 2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"log"
+	"os/exec"
+
+	"github.com/u-root/webboot/pkg/mountkexec"
+	"golang.org/x/sys/unix"
+)
+
+func main() {
+	var tests = []struct {
+		name string
+		iso  string
+		dir  string
+		err  string
+	}{
+		{"Empty Directory Name", "testIso", "", "error making mount directory:mkdir : no such file or directory"},
+		{"Non-existent Iso", "non-existentfile", "/tmp/mountDir", "error setting loop device:open non-existentfile: no such file or directory"},
+		{"GoodTest", "testIso", "/tmp/mountDir", ""},
+	}
+
+	for _, test := range tests {
+		log.Print(test.name)
+		if err := mountkexec.MountISO(test.iso, test.dir); err != nil {
+			log.Print(err)
+		} else {
+			o, err := exec.Command("ls", "/tmp/mountDir").CombinedOutput()
+			if err != nil {
+				log.Fatalf("ls failed: error %v", err)
+			}
+			log.Printf("%v", string(o))
+		}
+	}
+
+	unix.Reboot(unix.LINUX_REBOOT_CMD_POWER_OFF)
+}
--- a/pkg/mountkexec/mountkexec.go
+++ b/pkg/mountkexec/mountkexec.go
@@ -1,0 +1,34 @@
+// Copyright 2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package mountkexec
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/u-root/u-root/pkg/loop"
+	"github.com/u-root/u-root/pkg/mount"
+	"golang.org/x/sys/unix"
+)
+
+// MountISO creates a directory and mounts an input iso on the directory.
+func MountISO(isoPath, mountDir string) error {
+	if err := os.MkdirAll(mountDir, 777); err != nil {
+		return fmt.Errorf("error making mount directory:%v", err)
+	}
+	loopDevice, err := loop.FindDevice()
+	if err != nil {
+		return fmt.Errorf("error finding loop device:%v", err)
+	}
+	if err := loop.SetFile(loopDevice, isoPath); err != nil {
+		return fmt.Errorf("error setting loop device:%v", err)
+	}
+	var flags uintptr
+	flags |= unix.MS_RDONLY
+	if err = mount.Mount(loopDevice, mountDir, "iso9660", "", flags); err != nil {
+		return fmt.Errorf("error mounting ISO:%v", err)
+	}
+	return nil
+}
--- a/pkg/tlog/tlog.go
+++ b/pkg/tlog/tlog.go
@@ -1,0 +1,24 @@
+package tlog
+
+// Copyright 2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+import (
+	"testing"
+)
+
+//Testing implements testing interface
+type Testing struct {
+	Test *testing.T
+}
+
+//Print prints a input string
+func (t Testing) Print(v ...interface{}) {
+	t.Test.Log(v...)
+}
+
+//Printf prints a formated string
+func (t Testing) Printf(format string, v ...interface{}) {
+	t.Test.Logf(format, v...)
+}
--- a/pkg/webboot/webboot.go
+++ b/pkg/webboot/webboot.go
@@ -1,0 +1,9 @@
+package webboot
+
+//Distro defines an operating system distribution
+type Distro struct {
+	Kernel       string
+	Initrd       string
+	Cmdline      string
+	DownloadLink string
+}
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -1,4 +1,4 @@
-// Copyright 2013-2019 the u-root Authors. All rights reserved
+// Copyright 2019 the u-root Authors. All rights reserved
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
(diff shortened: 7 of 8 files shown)


webboot/webboot.go
```diff
@@ -2,15 +2,55 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// Get the time the machine has been up
+// Synopsis:
+//     webboot
+// Copyright 2013-2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
```
upatel99: The comments show up twice. Remove the repeated comments

louismurerwa: Done

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/u-root/u-root/pkg/sh"
+	webboot "github.com/u-root/webboot/pkg/mountkexec"
+	"golang.org/x/sys/unix"
+)
+
+// Mount a vfat volume and kexec the kernel within.
+func main() {
+
+	if err := webboot.MountISO("/dev/sda1", ""); err != nil {
```
rminnich: why is it /dev/sda1?

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,35 @@
...
+		log.Printf(err.Error())
+	}
+
+	if err := webboot.MountISO("/dev/sd", "mountDir"); err != nil {
```
rminnich: why is it /dev/sd? Is this only ever running in qemu?
If it works for now that is ok and we can change it later.

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,35 @@
...
+		log.Printf(err.Error())
+	}
+
+	if err := webboot.MountISO("/dev/sda1", "mountDir"); err != nil {
+		log.Fatal(err)
+	}
+
+	if err := sh.Run("ls", "mountDir/"); err != nil {
+		log.Fatal(err)
```
rminnich: This is not enough. I recommend you change it:
if o, err := exec.Command(""ls", "mountDir").CombinedOutput(); err != nil {
log.Fatalf("ls failed: %s, error %v", string(o), err)
}

pkg/mountkexec/mountkexec.go
```diff
@@ -0,0 +1,34 @@
...
+	loopDevice, err := loop.FindDevice()
+	if err != nil {
+		return fmt.Errorf("error finding loop device:%v", err)
+	}
+	if err := loop.SetFile(loopDevice, isoPath); err != nil {
+		return fmt.Errorf("error setting loop device:%v", err)
+	}
+	var flags uintptr
+	flags |= unix.MS_RDONLY
+	if err = mount.Mount(loopDevice, mountDir, "vfat", "", flags); err != nil {
```
rminnich: vfat?

webboot/webboot.go
```diff
@@ -8,9 +9,42 @@
...
+	initrd       string
+	cmdline      string
+	downloadLink string
+}
+
+var (
+	cmd       = flag.String("cmd", "", "Command Line")
+	mountDir  = flag.String("dir", "mountDir", "The mount point of the ISO")
+	osSystems = map[string]distro{
+		"TinyCore": distro{"/boot/vmliniz64 ", "/boot/initrd", "--reuse-cmdline", "https:louis.com"},
```
rminnich: vmlinuz64 right?

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: dev/sda1 is a vfat file created in intergration,go.I am using it to test my mountISO since I dont have an ISO yet.I will replace it with an actucal iso once I make one.

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: Here I am just trying to mount a non-existent file .This will return an error since the file is none existent hence the test passes.I will change the name to one that clearly shows what I am doing.

webboot/webboot.go
louismurerwa: Yes .

pkg/mountkexec/mountkexec.go
louismurerwa: This is the formart I am testing with for now.I will change it once I get an ISO

integration/mountkexec_test.go
```diff
@@ -0,0 +1,38 @@
...
+	q, cleanup := QEMUTest(t, &Options{
+		Cmds: []string{
+			"github.com/u-root/webboot/integration/testcmd/mountkexec/uinit",
+			"github.com/u-root/u-root/cmds/core/init",
+			"github.com/u-root/u-root/cmds/core/ls",
+		},
+	})
+	defer cleanup()
+
+	if err := q.Expect("error making mount directory:mkdir : no such file or directory"); err != nil {
```
rminnich: since this is a fixed set of strings would it be possible to do a for loop over a p[]string? you have lots of repeated code here.

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Mount a vfat volume and kexec the kernel within.
+func main() {
+	// Test : Mount on non-existent directoty
+	if err := mountkexec.MountISO("/dev/sda1", ""); err != nil {
+		log.Printf(err.Error())
+	}
+	// Test : Mount non-existent iso
+	if err := mountkexec.MountISO("non-existent file", "mountDir"); err != nil {
+		log.Printf(err.Error())
```
rminnich: log.Print(err)?

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	// Test : Mount on existing dir and an existing iso
+	if err := mountkexec.MountISO("/dev/sda1", "mountDir"); err != nil {
+		log.Fatal(err)
```
rminnich: this could also be a table, instead of blocks of repeated code.

webboot/webboot.go
```diff
@@ -8,9 +8,42 @@
...
+		"TinyCore": distro{"/boot/vmlinuz64 ", "/boot/initrd", "--reuse-cmdline", "https:louis.com"},
+	}
 )
 
 func main() {
-	fmt.Println("weboot up")
+
+	flag.Parse()
+	if len(flag.Args()) < 1 {
+		log.Fatal("error:pass in a valid operating system")
```
rminnich: could you, here, print out the list of possible names? You can iterate over your osSsystems map above.

webboot/webboot.go
```diff
@@ -8,9 +8,42 @@
...
+		os.Exit(1)
```
rminnich: you don't need an os.Exit here. log.Fatal exits.
in general you should provide a usage() function and call that for any errors.

webboot/webboot.go
```diff
@@ -8,9 +8,42 @@
...
+	}
+
+	osname := flag.Args()[0]
+	mountPath := *mountDir
```
rminnich: just use *mountDir below, no need for this extra variable.

webboot/webboot.go
```diff
@@ -8,9 +8,42 @@
...
+
+	if err := mountkexec.MountISO(osname, mountPath); err != nil {
+		log.Fatalf("error in mountISO:%v", err)
+		os.Exit(1)
```
rminnich: don 't need this os.Exit()

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	//Test : Check files in mounted iso
+	o, err := exec.Command("ls", "mountDir/").CombinedOutput()
+	if err != nil {
+		log.Fatalf("ls failed: error %v", err)
+	} else {
```
rminnich: you don't need the else. log.Fatalf will exit.

louismurerwa: I am using the else to make sure that the out put from running ls is returned to mountkexec_test.go.The results from running the command ls are stored in the object which I am printing in that else statement .

webboot/webboot.go
louismurerwa: Yes I agree also since log.Fatalf exists also.

webboot/webboot.go
louismurerwa: Done

webboot/webboot.go
louismurerwa: How do I implement this usage function?

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: Done

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: Done.

integration/mountkexec_test.go
louismurerwa: Done.This is more efficient.

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: Done

webboot/webboot.go
louismurerwa: Done .I like this idea.It helps a user figure out which operating systems are supported faster.

pkg/mountkexec/mountkexec.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/u-root/u-root/pkg/loop"
+	"github.com/u-root/u-root/pkg/mount"
+	"golang.org/x/sys/unix"
+)
+
+// MountISO runs a command with mount,unix and loop
```
rminnich: This comment needs more information. What does it do? 
Looks like mount the iso path via the loopback device? It does not appear to run a command, though.
Can you make this a bit more informative?

pkg/webboot/webboot.go
```diff
@@ -0,0 +1,9 @@
+package webboot
+
+//Distro struct defines an operating system distrubution
```
rminnich: you can remove the word struct from this comment. Maybe someday it gets changed to be an interface, and the comment would still be accurate as long as the word struct is not there.
distribution, not distrubution

webboot/webboot.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2013-2019 the u-root Authors. All rights reserved
+// Copyright 2019 the webboot Authors. All rights reserved
 // Use of this source code is governed by a BSD-style
```
rminnich: we had a discussion about this, we decided to leave it as u-root Authors

integration/testcmd/mountkexec/uinit/mountkexec.go
rminnich: The log print is fine. But since the Fatalf exits, you can remove 
} else {
and just have the log.Printf()

webboot/webboot.go
```diff
@@ -8,9 +8,42 @@
...
+
+	flag.Parse()
+	if len(flag.Args()) < 1 {
+		log.Fatal("error:pass in an operating system")
+	}
+
+	osname := flag.Args()[0]
+
+	if _, ok := osSystems[osname]; !ok {
+		log.Printf("error operating system is not supported")
```
rminnich: I'm wondering why you have two log.Printf instead of one?

webboot/webboot.go
louismurerwa: Done.

pkg/webboot/webboot.go
louismurerwa: Done.

webboot/webboot.go
louismurerwa: I wanted to make sure the error output is clear and easily readable by the user.

pkg/mountkexec/mountkexec.go
louismurerwa: Done.

buildimage.go
```diff
@@ -1,3 +1,6 @@
+// Copyright 2018 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
rminnich: 2019

integration/testcmd/mountkexec/uinit/mountkexec.go
rminnich: did you this note? you can replace
} else {
		log.Printf("%v", string(o))
	}
with
		log.Printf("%v", string(o))

buildimage.go
louismurerwa: Done

integration/testcmd/mountkexec/uinit/mountkexec.go
```diff
@@ -0,0 +1,37 @@
...
+		{"non-existentfile", "/tmp/mountDir"},
+		{"testIso", "/tmp/mountDir"},
+	}
+
+	for _, test := range tests {
+		if err := mountkexec.MountISO(test.iso, test.dir); err != nil {
+			log.Print(err)
+		}
+	}
+	o, err := exec.Command("ls", "/tmp/mountDir").CombinedOutput()
```
GanShun: This assumes that the last mount works. Put this inside the for loop instead, and I suggest you name the subtests as you run them, and in the expect, check for the name as part of the string to be expected. You want to know that the "non-existentfile" test really was the test that produced "error setting loop device:open non-existentfile: no such file or directory", and not another test. Don't just rely on the order of the tests.

integration/mountkexec_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	if err = exec.Command("genisoimage", "-o", "/tmp/tempdata.iso", tmpDir).Run(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// TestMountKexec runs an init which mounts a filesystem and kexecs a kernel.
+func TestKexecMount(t *testing.T) {
```
GanShun: This test doesn't actually kexec?

louismurerwa: Yes but I am planning to extend it once I implement the kexec function.

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: Done

integration/testcmd/mountkexec/uinit/mountkexec.go
louismurerwa: Done.

integration/mountkexec_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"testing"
+
+	"github.com/u-root/u-root/pkg/cp"
+	"github.com/u-root/u-root/pkg/golang"
+	"github.com/u-root/u-root/pkg/uroot"
+	"github.com/u-root/u-root/pkg/uroot/builder"
+	"github.com/u-root/u-root/pkg/uroot/initramfs"
+)
+
+// Generates an iso containing a kernel and initramfs to be used for testing mount and kexec
```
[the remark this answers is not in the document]
louismurerwa: Done

integration/mountkexec_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// OutputFile
+	newOutputFile := filepath.Join(tmpDir, "initiso.cpio")
+	w, err := initramfs.CPIO.OpenWriter(log.New(os.Stderr, "", 0), newOutputFile, "", "")
+	if err != nil {
+		return err
+	}
+
+	var cmds []string
+	cmds = append(cmds, "github.com/u-root/u-root/cmds/*")
```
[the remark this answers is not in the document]
louismurerwa: Done

integration/mountkexec_test.go
```diff
@@ -0,0 +1,105 @@
...
+		return err
+	}
+	kernel := filepath.Join(tmpDir, "kernel")
+	if err := cp.Copy(os.Getenv("UROOT_KERNEL"), kernel); err != nil {
+		return err
+	}
+
+	// OutputFile
+	newOutputFile := filepath.Join(tmpDir, "initiso.cpio")
+	w, err := initramfs.CPIO.OpenWriter(log.New(os.Stderr, "", 0), newOutputFile, "", "")
```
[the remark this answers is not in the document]
louismurerwa: Hey Ron.I don't understand what I am supposed to be doing here.First of all,I don't think I can use t.Log here because genISO doesn't take in a testing parameter,so I don't have access to t.Log.Also why should we never use log?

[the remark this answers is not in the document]
rminnich: damn github, I was seeing an older version, even after a refresh.

Anyway, I'm gonna let you work this one out. The problem: can't use log.New() here. So, you need to see how to fix it.

integration/mountkexec_test.go
```diff
@@ -0,0 +1,107 @@
...
+		TempDir:      tmpDir,
+		BaseArchive:  uroot.DefaultRamfs.Reader(),
+		OutputFile:   w,
+		DefaultShell: "elvish",
+	}
+	if err := uroot.CreateInitramfs(log.New(os.Stderr, "", 0), opts); err != nil {
+		return err
+	}
+
+	if err = exec.Command("genisoimage", "-o", "/tmp/tempdata.iso", tmpDir).Run(); err != nil {
```
rminnich: use CombinedOutput, not Run, so you can have some control over where output goes. This is a test and it is not allowed to generate lots of output in an uncontrolled way.

integration/mountkexec_test.go
```diff
@@ -0,0 +1,111 @@
...
+		DefaultShell: "elvish",
+	}
+	var tlogger *tlog.Testing
+	tlogger.Test = t
+	if err := uroot.CreateInitramfs(tlogger, opts); err != nil {
+		return err
+	}
+	_, err = exec.Command("genisoimage", "-o", "/tmp/tempdata.iso", tmpDir).CombinedOutput()
+	if err != nil {
+		return err
```
rminnich: o, err := exec.Command(...)
return fmt.Printf("genisoimage command: got (%v, %v), want nil", o, err)

rminnich: return fmt.Errorf("%v: %v", o, err)

louismurerwa: Done.

integration/mountkexec_test.go
louismurerwa: Done

integration/mountkexec_test.go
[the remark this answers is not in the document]
rminnich: did this get changed?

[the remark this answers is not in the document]
rminnich: ah, did you want tlogger here too?

pkg/mountkexec/mountkexec.go
```diff
@@ -0,0 +1,34 @@
...
+		return fmt.Errorf("error making mount directory:%v", err)
+	}
+	loopDevice, err := loop.FindDevice()
+	if err != nil {
+		return fmt.Errorf("error finding loop device:%v", err)
+	}
+	if err := loop.SetFile(loopDevice, isoPath); err != nil {
+		return fmt.Errorf("error setting loop device:%v", err)
+	}
+	var flags uintptr
```
rminnich: so maybe next time around, you can do this
flags := unix.MS_RDONLY

pkg/tlog/tlog.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Copyright 2019 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+import (
+	"testing"
+)
+
+//Testing implements testing interface
```
rminnich: does it? I thought it implemented log.Logger?

webboot/webboot.go
```diff
@@ -12,20 +12,46 @@ import (
...
 	dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
-	log.Println("weboot up")
+
+	if len(flag.Args()) < 1 {
+		log.Fatal("error:pass in an operating system")
+	}
+
+	osname := flag.Args()[0]
+
+	if _, ok := osSystems[osname]; !ok {
```
rminnich: you'll need to talk to urvisha about this, how do we work when people give you a valid URL.

louismurerwa: Urvisha got it working on her branch and we should be able to get it in on the next PR.

pkg/tlog/tlog.go
louismurerwa: It implements the Print and Printf which are required to satisfy the log.Logger interface.Hence it does implement it indirectly.

pkg/mountkexec/mountkexec.go
louismurerwa: I am making this change in my next PR

PR #13, opened 2019-08-13
Add iwconfig and iwlist to Webboot initrd

upatel99:
Webboot now includes the WIFI feature. In order for WIFI
to work, it needs to append the iwconfig and iwlist when building
its initramfs to set parameters of network interface specific to
wireless operation and to see information about the interfaces.


GanShun:
initrd, not intird

PR #16, opened 2019-08-14
Add mykernel, an option to boot tinycore initramfs with our bzImage kernel.

louismurerwa:

### DIFF
--- a/buildimage.go
+++ b/buildimage.go
@@ -12,7 +12,8 @@ import (
 )
 
 var (
-	debug   = func(string, ...interface{}) {}
+	debug = func(string, ...interface{}) {}
+
 	verbose = flag.Bool("v", true, "verbose debugging output")
 	uroot   = flag.String("u", "", "options for u-root")
 	cmds    = flag.String("c", "core", "u-root commands to build into the image")
@@ -35,6 +36,7 @@ func main() {
 		{"sudo", "apt", "install", "wireless-tools"},
 		append(append([]string{"go", "run", "github.com/u-root/u-root/.", "-files", "/sbin/iwconfig:bin/iwconfig", "-files", "/sbin/iwlist:bin/iwlist"}, strings.Fields(*uroot)...), *cmds, *wcmds, *ncmds),
 	}
+
 	for _, cmd := range commands {
 		debug("Run %v", cmd)
 		c := exec.Command(cmd[0], cmd[1:]...)
--- a/pkg/mountkexec/mountkexec.go
+++ b/pkg/mountkexec/mountkexec.go
@@ -7,6 +7,7 @@ package mountkexec
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/u-root/u-root/pkg/boot"
 	"github.com/u-root/u-root/pkg/kexec"
@@ -39,17 +40,22 @@ func MountISO(isoPath, mountDir string) error {
 }
 
 // KexecISO boots up a new kernel and initramfs
-func KexecISO(opp *webboot.Distro, path string) error {
+func KexecISO(opp *webboot.Distro, dir string) error {
 	var image boot.OSImage
-	if err := multiboot.Probe(path + opp.Kernel); err == nil {
+	kernelPath := opp.Kernel
+	if !filepath.IsAbs(kernelPath) {
+		kernelPath = filepath.Join(dir, kernelPath)
+	}
+
+	if err := multiboot.Probe(kernelPath); err == nil {
 		image = &boot.MultibootImage{
-			Path:    opp.Kernel,
+			Path:    kernelPath,
 			Cmdline: opp.Cmdline,
 		}
 	} else {
 		image = &boot.LinuxImage{
-			Kernel:  uio.NewLazyFile(path + opp.Kernel),
-			Initrd:  uio.NewLazyFile(path + opp.Initrd),
+			Kernel:  uio.NewLazyFile(kernelPath),
+			Initrd:  uio.NewLazyFile(dir + opp.Initrd),
 			Cmdline: opp.Cmdline,
 		}
 	}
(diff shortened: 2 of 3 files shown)


buildimage.go
```diff
@@ -32,7 +32,11 @@ func main() {
 		{"go", "get", "-u", "github.com/u-root/u-root"},
 		{"go", "get", "-d", "-v", "-u", "github.com/u-root/NiChrome/..."},
 		{"sudo", "apt", "install", "wireless-tools"},
-		{"go", "run", "github.com/u-root/u-root/.", "-files", "/sbin/iwconfig:bin/iwconfig", "-files", "/sbin/iwlist:bin/iwlist", *uroot, *cmds, *wcmds, *ncmds},
+		{"go", "run", "github.com/u-root/u-root/.", "-files", "bzImage:bzImage", "-files", "/sbin/iwconfig:/bin/iwconfig", "-files", "/sbin/iwlist:/bin/iwlist", *uroot, *cmds, *wcmds, *ncmds},
```
upatel99: this test is failing because you are trying to put the bzImage in the initramfs. However, the dependency test fails because it cannot find the bzImage to put in... Perhaps, passing the kernel with these files in the commit will help pass the test? What do you think, Ron and Gan? @rminnich @GanShun

rminnich: The big problem is that we are putting -files -bzImage:bzImage into this tool when there may not be a bzImage file.

I think maybe we should have a way to pass "extra" files via the command line. That's what I was trying to do yesterday but failed.

louismurerwa: Done

webboot/webboot.go
```diff
@@ -32,10 +32,11 @@ var (
 	verbose  = flag.Bool("verbose", false, "Verbose output")
 	ipv4     = flag.Bool("ipv4", true, "use IPV4")
 	ipv6     = flag.Bool("ipv6", true, "use IPV6")
+	dryrun   = flag.Bool("dryrun", false, "Do not do the kexec")
 	bookmark = map[string]*webboot.Distro{
-		"tinycore": &webboot.Distro{"/boot/vmlinuz64", "/boot/corepure64.gz", "console=tty1", "http://tinycorelinux.net/10.x/x86_64/release/CorePure64-10.1.iso"},
+		"tinycore": &webboot.Distro{"boot/vmlinuz64", "/boot/corepure64.gz", "console=tty0", "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-current.iso"},
+		"mykernel": &webboot.Distro{"/bzImage", "/boot/corepure64.gz", "console=tty0", "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-current.iso"},
```
upatel99: name this something more meaningful than "mykernel" :)

buildimage.go
```diff
@@ -32,7 +33,11 @@ func main() {
 		{"go", "get", "-u", "github.com/u-root/u-root"},
 		{"go", "get", "-d", "-v", "-u", "github.com/u-root/NiChrome/..."},
 		{"sudo", "apt", "install", "wireless-tools"},
-		{"go", "run", "github.com/u-root/u-root/.", "-files", "/sbin/iwconfig:bin/iwconfig", "-files", "/sbin/iwlist:bin/iwlist", *uroot, *cmds, *wcmds, *ncmds},
+		{"go", "run", "github.com/u-root/u-root/.", "-files", "/sbin/iwconfig:/bin/iwconfig", "-files", "/sbin/iwlist:/bin/iwlist", *uroot, *cmds, *wcmds, *ncmds},
+	}
+	for _, cmd := range commands {
+		debug("Run %v", cmd)
```
upatel99: why is there two for loops ranging over the same thing and performing the same acts ?

webboot/webboot.go
louismurerwa: You have to talk to Ron about that.

louismurerwa: "mykernel" clearly tells you that the program is booting using the users kernel and not the one packaged originally into the iso.Can you suggest some other names which might work here?

buildimage.go
louismurerwa: The other loop came from merging your edits into my branch.I removed the second loop.

pkg/mountkexec/mountkexec.go
```diff
@@ -41,14 +42,19 @@ func MountISO(isoPath, mountDir string) error {
 // KexecISO boots up a new kernel and initramfs
 func KexecISO(opp *webboot.Distro, path string) error {
 	var image boot.OSImage
+	kernelPath := opp.Kernel
+	if !filepath.IsAbs(kernelPath) {
+		kernelPath = filepath.Join(path, "/", opp.Kernel)
```
rminnich: you do not need the "/". filepath.Join will put it in for you.

Since you are going to change this anyway, could you change the argument name from 'path' to 'dir'? It's a bit confusing for me.

louismurerwa: Done.

pkg/mountkexec/mountkexec.go
```diff
@@ -39,17 +40,22 @@ func MountISO(isoPath, mountDir string) error {
 }
 
 // KexecISO boots up a new kernel and initramfs
-func KexecISO(opp *webboot.Distro, path string) error {
+func KexecISO(opp *webboot.Distro, dir string) error {
 	var image boot.OSImage
-	if err := multiboot.Probe(path + opp.Kernel); err == nil {
+	kernelPath := opp.Kernel
+	if !filepath.IsAbs(kernelPath) {
+		kernelPath = filepath.Join(dir, opp.Kernel)
```
rminnich: you can change this later but you could consider using kernelPath instead of opp.Kernel. No big deal.

louismurerwa: Thats true .Let me make the edit.

PR #17, opened 2019-08-14
Clean up, add comments, remove a directory flag

upatel99:
Delete some confusing and misleading comments, and flag descriptions.
Add package comments, explaining webboot's implementations.

Remove mountDir flag and store the contents in a temporary mount directory.
Temporary directories are much safer to use because before two or more people
could run into each other if they run the same command, giving the same
directory name, but now, people get unique temporary directories that is
not already utilized.

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -1,16 +1,27 @@
 // Copyright 2019 the u-root Authors. All rights reserved
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
-
-// Get the time the machine has been up
+//
 // Synopsis:
-//     webboot
+//     webboot [OPTIONS...] URL or name of bookmark
+//
+// Options:
+//	-cmd: Command line parameters to the second kernel
+//	-mountDir: Mount point
+//	-ifName: Name of the interface
+//	-timeout: Lease timeout in seconds
+//	-retry: Number of DHCP renewals before exiting
+//	-verbose:  Verbose output
+//	-ipv4: Use IPV4
+//	-ipv6: Use IPV6
+//	-dryrun: Do not do the kexec
 package main
 
 import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


webboot/webboot.go
```diff
@@ -1,10 +1,20 @@
 // Copyright 2019 the u-root Authors. All rights reserved
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
-
-// Get the time the machine has been up
+//
 // Synopsis:
-//     webboot
+//     webboot [OPTIONS...] URL or name of ISO
```
rminnich: name of bookmark? just wondering.

webboot/webboot.go
```diff
@@ -24,8 +34,8 @@ import (
 )
 
 var (
-	cmd      = flag.String("cmd", "", "Command Line")
-	mountDir = flag.String("dir", "/tmp/mountDir", "The mount point of the ISO")
+	cmd      = flag.String("cmd", "", "Command line parameters to the second kernel")
+	mountDir = flag.String("dir", "/tmp/mountDir", "The mount point")
```
rminnich: we should probably be using ioutil.TempDir and drop this option.
ioutil.TempDir is much safer to use anyway. Could you fold that into this PR? Or do another PR?

webboot/webboot.go
upatel99: Yes, good point. Name of bookmark flows better with the description. Name of bookmark are names of OS/ISOs, but yes, I definitely agree with changing the comment to include name of bookmark instead. Thanks for pointing that out :)!

webboot/webboot.go
upatel99: My only question is, this will allows less flexibility for the user, right? What if the user wants to mount it onto another existing directory? Should we allow them to do so, or should we just store it in a temp directory using ioutil.TempDir? I could fold that into this PR for sure though.

rminnich: I don't see any value in that and if two people run this at the same time, and they pick the same name, or if they use the default name, they're going to run into each other. That problem is why ioutil.TempDir exists -- so people can get a unique TempDir that somebody else won't use. So thinking about this, I'd say let's get rid of mountDIr and ONLY use ioutil.TempDir

upatel99: That is a good point. I will go ahead and delete the mountDir flag and just utilize the ioutil.TempDir instead

webboot/webboot.go
```diff
@@ -104,13 +115,12 @@ func name(URL string) (string, error) {
 }
 
 func usage() {
-	log.Printf("Usage: %s [ARGS] URL or name of ISO\n", os.Args[0])
+	log.Printf("Usage: %s [ARGS] URL or name of OS\n", os.Args[0])
```
rminnich: you can do a new PR but I just realized ... [ARGS] is the wrong word. it should be [flags]. An arg is the URL or the bookmark. The other things are flags.

upatel99: Okay,  I can handle this in another PR :)

PR #25, opened 2019-08-19
Download ISO and save it in pmem device

upatel99:
Store the ISO in pmem device that will be
carried across to the next kernel. This allows the next kernel to load
all of the packages for graphics and saves time from storing it
into a file for mounting when pmem device can be
used instead.

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -79,22 +79,6 @@ func linkOpen(URL string) (io.ReadCloser, error) {
 	return resp.Body, nil
 }
 
-// write copies the content from an io.readCloser to a named file with filename.
-func write(read io.ReadCloser, filename string) error {
-	w, err := os.Create(filename)
-	if err != nil {
-		return err
-	}
-
-	defer w.Close()
-	defer read.Close()
-
-	if _, err := io.Copy(w, read); err != nil {
-		return fmt.Errorf("Error copying %v: %v", filename, err)
-	}
-	return nil
-}
-
 // name takes a URL and generates a filename from it
 // For example, if a valid URL = http://tinycorelinux.net/10.x/x86_64/release/CorePure64-10.1.iso, then filename = CorePure64-10.1.iso
 // if the URL is empty or if the URL's Path ends in /, name returns a default index.html as the filename
@@ -148,14 +132,22 @@ func main() {
 
 	//Processes the URL to receive an io.ReadCloser, which holds the content of the downloaded file
 	log.Println("Retrieving the file...")
-	read, err := linkOpen(URL)
+	iso, err := linkOpen(URL)
 	if err != nil {
 		log.Fatal(err)
 	}
+	defer iso.Close()
 
-	err = write(read, filename)
+	// TODO: Find a persistent memory device large enough to store the content. If no blocks are available, error the user.
+	pmem, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)
 	if err != nil {
 		log.Fatal(err)
+	}
+	if _, err := io.Copy(pmem, iso); err != nil {
+		log.Fatalf("Error copying to persistent memory device: %v", err)
+	}
+	if err = pmem.Close(); err != nil {
+		log.Fatalf("Error closing /dev/pmem0: %v", err)
 	}
 
 	tmp, err := ioutil.TempDir("", "mnt")
--- a/webboot/webboot_test.go
+++ b/webboot/webboot_test.go
@@ -63,31 +63,6 @@ func TestName(t *testing.T) {
 
 }
 
-func TestWrite(t *testing.T) {
-	content := []byte("temporary file's content")
-	tmpDir, err := ioutil.TempDir("", "tmpDir")
-	if err != nil {
-		t.Fatalf("Failed to create a temporary directory")
-	}
-	tmpfile := filepath.Join(tmpDir, "tmpfile")
-	if err := ioutil.WriteFile(tmpfile, content, 0600); err != nil {
-		t.Fatal(err)
-	}
-
-	file, err := os.Open(tmpfile)
-	if err != nil {
-		t.Fatalf("Failed to open %v: got %v, want nil", tmpfile, err)
-	}
-
-	testfn := filepath.Join(tmpDir, "testfile")
-	err = write(file, testfn)
-
-	read, err := ioutil.ReadFile(testfn)
-	if !bytes.Equal(content, read) {
-		t.Fatalf("Failed to write %v: got %v, want %v", testfn, string(read), string(content))
-	}
-}
-
 func TestLinkOpen(t *testing.T) {
 	tests := []test{
 		{name: "TinyCore", linkOrName: "http://tinycorelinux.net/10.x/x86_64/release/CorePure64-10.1.iso", md5link: "http://tinycorelinux.net/10.x/x86_64/release/CorePure64-10.1.iso.md5.txt"},


webboot/webboot.go
```diff
@@ -79,18 +79,13 @@ func linkOpen(URL string) (io.ReadCloser, error) {
...
-func write(read io.ReadCloser, filename string) error {
-	w, err := os.Create(filename)
-	if err != nil {
-		return err
-	}
-
-	defer w.Close()
-	defer read.Close()
+// TODO: Find a persistent memory block large enough to store the content. If no blocks are available, error the user.
+// write copies the content from an io.readCloser to a persistent memory block.
```
rminnich: instead of persistent memory block, just call it persistent memory device.

webboot/webboot.go
```diff
@@ -79,18 +79,13 @@ func linkOpen(URL string) (io.ReadCloser, error) {
...
+func write(body []byte, pmem *os.File) error {
+	defer pmem.Close()
```
rminnich: put this defer at line 155, not here. The function should not close a file passed as a parameter.

webboot/webboot.go
```diff
@@ -79,18 +79,13 @@ func linkOpen(URL string) (io.ReadCloser, error) {
...
-	w, err := os.Create(filename)
-	if err != nil {
-		return err
-	}
-
-	defer w.Close()
-	defer read.Close()
+// TODO: Find a persistent memory block large enough to store the content. If no blocks are available, error the user.
+// write copies the content from an io.readCloser to a persistent memory block.
+func write(body []byte, pmem *os.File) error {
```
rminnich: you can remove this function completely.

webboot/webboot.go
```diff
@@ -153,7 +148,16 @@ func main() {
 		log.Fatal(err)
 	}
 
-	err = write(read, filename)
+	file, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	body, err := ioutil.ReadAll(read)
```
rminnich: do not read the body here.

webboot/webboot.go
```diff
@@ -153,7 +148,16 @@ func main() {
 		log.Fatal(err)
 	}
 
-	err = write(read, filename)
+	file, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
rminnich: just put an io.Copy(file, read)
here
also change the name of 
file, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)
pmem, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)

and change the name read to something a little nicer, like, iso, maybe?

webboot/webboot.go
```diff
@@ -153,7 +148,16 @@ func main() {
...
+	body, err := ioutil.ReadAll(read)
+	if err != nil {
```
rminnich: get rid of lines 156-163. you don't need them.

webboot/webboot.go
upatel99: Sounds good. I will call it persistent memory device instead.

webboot/webboot.go
upatel99: That is a good point. I will let main take care of it all since it is opening the file then it should be responsible for closing the file as well, so the logic takes consistent.

webboot/webboot.go
upatel99: I was thinking the same. Using the go packages for opening the file, and copying the contents over to another file eliminates having to use an entire function.

webboot/webboot.go
upatel99: I realized I do not need to read the body at all. I will get rid of this.

webboot/webboot.go
upatel99: Okay, I will change the variable "file" to "pmem," and "read" to a variable name such as "iso," so it is very clear to the audience.

webboot/webboot.go
upatel99: Sounds good! I realized I do not need this if I can just open the persistent memory device and copy the content over to it.

webboot/webboot.go
```diff
@@ -148,15 +132,25 @@ func main() {
...
-	err = write(read, filename)
+	// TODO: Find a persistent memory device large enough to store the content. If no blocks are available, error the user.
+	pmem, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)
 	if err != nil {
 		log.Fatal(err)
 	}
+	if _, err := io.Copy(pmem, iso); err != nil {
+		log.Fatalf("Error copying to persistent memory device: %v", err)
+	}
+	if err = iso.Close(); err != nil {
```
upatel99: I have a question here: what is a good practice for defering Close() and checking the error for Close() right after I am done utilizing it? I figured since I am done using the io.Reader here, that I can just check the error if I receive an error from closing here? However, I can also defer a Close() once main() is done executing..

webboot/webboot.go
```diff
@@ -148,15 +132,25 @@ func main() {
...
+		log.Fatalf("Error closing reader for %v: %v", URL, err)
+	}
+	if err = pmem.Close(); err != nil {
```
upatel99: I have a question here: what is a good practice for defering Close() and checking the error for Close() right after I am done utilizing it? I figured since I am done using the file here, that I can just check the error if I receive an error from closing here? However, I can also defer a Close() once main() is done executing..

webboot/webboot.go
```diff
@@ -148,15 +132,25 @@ func main() {
...
+		log.Fatalf("Error closing /dev/pmem0: %v", err)
```
upatel99: is it okay for me to error the user out letting them know that the error occurred closing /dev/pmem0 such as this?

webboot/webboot.go
```diff
@@ -148,15 +132,25 @@ func main() {
...
+	// TODO: Find a persistent memory device large enough to store the content. If no blocks are available, error the user.
+	pmem, err := os.OpenFile("/dev/pmem0", os.O_APPEND|os.O_WRONLY, 0600)
 	if err != nil {
 		log.Fatal(err)
 	}
+	if _, err := io.Copy(pmem, iso); err != nil {
+		log.Fatalf("Error copying to persistent memory device: %v", err)
+	}
+	if err = iso.Close(); err != nil {
+		log.Fatalf("Error closing reader for %v: %v", URL, err)
```
upatel99: I like this style of error message because this will let the user know that the reader for a URL was having trouble closing and listing the specific URL will let the user know which specific URL was having problem closing. However, I am open to suggestions for changing the printing style :)

webboot/webboot.go
```diff
@@ -148,15 +132,25 @@ func main() {
 
 	//Processes the URL to receive an io.ReadCloser, which holds the content of the downloaded file
 	log.Println("Retrieving the file...")
-	read, err := linkOpen(URL)
+	iso, err := linkOpen(URL)
 	if err != nil {
 		log.Fatal(err)
 	}
```
rminnich: defer iso.Close()

webboot/webboot.go
rminnich: yes. A close error can mean that there was an IO error of some sort. This is fine.

webboot/webboot.go
rminnich: generally, it's best to check the error from a Close on a file you are writing. When reading, defer'ing a close is fine.

webboot/webboot.go
rminnich: defer iso.Close()
is fine here.

upatel99: Okay, sounds good.  I did have a question when one should defer a Close() and when one should check an error, but after your explanation in one of the comments below, it makes more sense now :) thanks!

upatel99: Sounds good :)!

webboot/webboot.go
upatel99: Okay, sounds good :) thank you

webboot/webboot.go
upatel99: Oh okay, that makes more sense. Since iso is only read from, then I can defer it, but since I am writing to pmem0, then I should check if an error is provided when trying to close it. Thank you for answering my question :)

PR #26, opened 2019-08-19
Built in the wifi feature

louismurerwa:
Added the -wifi flag to allow a user to select their wifi.
Built in GoogleGuest as the default wifi.

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"path"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


webboot/webboot.go
```diff
@@ -43,6 +44,7 @@ var (
 	ipv4     = flag.Bool("ipv4", true, "use IPV4")
 	ipv6     = flag.Bool("ipv6", true, "use IPV6")
 	dryrun   = flag.Bool("dryrun", false, "Do not do the kexec")
+	wifi     = flag.String("wifi", "GoogleGuest", "Wifi name")
```
rminnich: suggestion: 
Change 
"Wifi name"
 to
"ESSID"
it's something people will get right away
you can also change the flag name from
"wifi"
to 
"essid"

webboot/webboot.go
```diff
@@ -117,12 +118,24 @@ func usage() {
 	os.Exit(1)
 }
 
+func setupWifi(wifi string) error {
```
rminnich: The function name is great, but change the variable wifi to essid

webboot/webboot.go
```diff
@@ -117,12 +118,24 @@ func usage() {
...
+	o, err := exec.Command("iwconfig", "wlan0", "essid", wifi).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error setting up wifi:%v", err)
+	}
+	log.Printf("Connected to Wifi Succesfully %v", string(o))
```
rminnich: what if the command ran but iwconfig had an error anyway? 
Maybe you could do
log.Printf("setupWifi: iwconfig returned %v", string(o))

upatel99: Should there be a print in the function anyway? Wouldn't this be an issue for testing or other things? Instead, could there not be a print statement in main() after the function runs successfully? By then, wifi would be successfully connected and all of the errors would be handled before the print message in main()

webboot/webboot.go
louismurerwa: Done.I didn't know WiFi was also called essid. Learnt something new today :)

webboot/webboot.go
louismurerwa: Done

webboot/webboot.go
louismurerwa: Yeah I think there should be a print in here because we need to make sure we received the correct message from running iwconfig.Any output from iwconfig should be printed out to enable us to debug in case of an error or connection failure.

PR #30, opened 2019-08-24
Add more wifi firmware so x230 works

rminnich:
Signed-off-by: Ronald G Minnich <rminnich@gmail.com>


rminnich:
there are still some other weird problems.

This fails on my x230 now in dynamic mode. Something is really messed up with vendoring.

But this change won't affect any of that, it just adds more firmware files.

built on
Ubuntu 18.04.3 LTS

PR #44, opened 2020-01-19
webboot: add Ubuntu and Arch Linux

orangecms:
this also reenables the original `tinycore` command

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -14,7 +14,7 @@
 //	-ipv4: Use IPV4
 //	-ipv6: Use IPV6
 //	-dryrun: Do not do the kexec
-//	-wifi: [essid [WPA [password]]] 
+//	-wifi: [essid [WPA [password]]]
 package main
 
 import (
@@ -49,7 +49,7 @@ var (
 	ipv4     = flag.Bool("ipv4", true, "use IPV4")
 	ipv6     = flag.Bool("ipv6", true, "use IPV6")
 	dryrun   = flag.Bool("dryrun", false, "Do not do the kexec")
-	wifi    = flag.String("wifi", "GoogleGuest", "[essid [WPA [password]]]")
+	wifi     = flag.String("wifi", "GoogleGuest", "[essid [WPA [password]]]")
 	bookmark = map[string]*webboot.Distro{
 		// TODO: Fix webboot to process the tinycore's kernel and initrd to boot from instead of using our customized kernel
 		"tinycore": &webboot.Distro{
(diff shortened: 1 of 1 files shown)


webboot/webboot.go
```diff
@@ -45,10 +50,49 @@ var (
 	wifi    = flag.String("wifi", "GoogleGuest", "[essid [WPA [password]]]")
 	bookmark = map[string]*webboot.Distro{
 		// TODO: Fix webboot to process the tinycore's kernel and initrd to boot from instead of using our customized kernel
-		// "tinycore": &webboot.Distro{"boot/vmlinuz64", "/boot/corepure64.gz", "console=tty0", "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"},
-		"Tinycore": &webboot.Distro{"/bzImage", "/boot/corepure64.gz", "memmap=4G!4G console=tty1 root=/dev/pmem0 loglevel=3 cde waitusb=5 vga=791", "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"},
+		"tinycore": &webboot.Distro{
+			"boot/vmlinuz64",
```
bjakobson: //kernel from TCs' ISO

bjakobson:
there's a few errors. when I copied the command that failed: 

go run github.com/u-root/u-root/. -files /sbin/iwconfig -files /sbin/iwlist -files /sbin/wpa_supplicant -files /sbin/wpa_action -files /sbin/wpa_cli -files /usr/bin/wpa_passphrase -build=bb all github.com/u-root/webboot/webboot/. github.com/u-root/NiChrome/cmds/wifi

I had to export GOMODULE111=off, then I am out of memory. I'm not sure if either of these is what caused azure to fail...

webboot/webboot.go
```diff
@@ -33,6 +33,11 @@ import (
 	"github.com/u-root/webboot/pkg/webboot"
 )
 
+const tcUrl = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
```
rminnich: if you want:
const (
blah = ""
blee = ""

and so on
)
no need to repeat const. I will approve and leave it up to you.

orangecms:
Correcy @rminnich - no distro ISO works upfront because none of them have pmem support, so we can only boot them with the customized kernels, just like with TinyCore, hence the second commands, following the pattern.

webboot/webboot.go
orangecms: Done, looks much nicer indeed. :)

PR #47, opened 2020-01-22
Kexec mount

bjakobson:
pkg/mountkexec/mountkexec.go broke after a change in u-root's multiboot, causing webboot to not be able to have it's initramfs built.

go.mod and go.sum has not been updated, and was automatically updated.

Signed-off-by: Brandon Jakobson <brandonjakobson@gmail.com>


pkg/mountkexec/mountkexec.go
```diff
@@ -52,13 +52,19 @@ func MountISOPmem(pmem, mountDir string) error {
 func KexecISO(opp *webboot.Distro, dir string) error {
 	var image boot.OSImage
 	kernelPath := opp.Kernel
+
+	kernel, err := os.Open(kernelPath)
+
+	if err != nil {
+		fmt.Errorf("Error: %v", err)
```
rminnich: Get rid of the Error: OR make it a lot more meaningful:
KexecISO:

webboot/.bb/webboot.go
```diff
@@ -0,0 +1,218 @@
+// Copyright 2019 the u-root Authors. All rights reserved
```
rminnich: this file should not be here. 
git reset HEAD^ webboot/.bb/webboot.go

bjakobson: Yeah i don’t know what happened, it added a file called .bb, let me make some changes

pkg/mountkexec/mountkexec.go
bjakobson: So i guess no error necessary, because multiboot.Probe won’t work if kernel doesn’t.

pkg/mountkexec/mountkexec.go
```diff
@@ -52,13 +52,16 @@ func MountISOPmem(pmem, mountDir string) error {
 func KexecISO(opp *webboot.Distro, dir string) error {
 	var image boot.OSImage
 	kernelPath := opp.Kernel
+
```
rminnich: This is in the wrong place. It has to be AFTER line 62

pkg/mountkexec/mountkexec.go
```diff
@@ -52,13 +52,16 @@ func MountISOPmem(pmem, mountDir string) error {
...
+	//kernel can be used in multiboot.Probe
+	kernel, _ := os.Open(kernelPath)
```
rminnich: you need to check the error on open. Never ignore errors.

PR #49, opened 2020-01-25
    Add options for build and runtime for more flexibility

rminnich:
Add options for build and runtime for more flexibility
    
    buildimage options:
    -bzImage add optional bzImage
    -iso add optional iso
    
    webboot:
    -wifi default to "" now, and if it is "", don't try to set up wifi
    
    new "local" distro, where the bzImage and iso are already on the initramfs
    
    linkOpen now supports the file:// schema (Go http package does not, BOOO!)
    so that you can just load a local file rather than having to get one.
    
    We can now do testing in other ways.
    
    buildimage -bzImage path/to/bzImage -iso path/to/tinycore.iso etc. etc.
    the bzImage will be placed in the image as bzImage, and the iso will
    be placed as iso
    
    if you specify local as the webboot distro, e.g.
    webboot -dhcp4=false -dhcp6-false local
    
    then it will use /bzImage as kernel, and /iso as iso, and away you go.
    
    This will greatly speed up debugging.
    
    Signed-off-by: Ronald G. Minnich <rminnich@gmail.com>

### DIFF
--- a/buildimage.go
+++ b/buildimage.go
@@ -15,10 +15,12 @@ var (
 	debug = func(string, ...interface{}) {}
 
 	verbose = flag.Bool("v", true, "verbose debugging output")
-	uroot   = flag.String("u", "-build=bb", "options for u-root")
-	cmds    = flag.String("c", "core", "u-root commands to build into the image")
+	uroot   = flag.String("u", "", "options for u-root")
+	cmds    = flag.String("c", "all", "u-root commands to build into the image")
 	wcmds   = flag.String("w", "github.com/u-root/webboot/webboot/.", "webboot commands to build into the image")
 	ncmds   = flag.String("n", "github.com/u-root/NiChrome/cmds/wifi", "NiChrome commands to build into the image")
+	bzImage = flag.String("bzImage", "", "Optional bzImage to embed in the initramfs")
+	iso     = flag.String("iso", "", "Optional iso (e.g. tinycore.iso) to embed in the initramfs")
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


rminnich:
I'm finding this way handier to debug ...

buildimage.go
```diff
@@ -34,23 +36,31 @@ func init() {
...
+		"go", "run", "github.com/u-root/u-root/.",
+		"-files", extraBinMust("iwconfig"),
+		"-files", extraBinMust("iwlist"),
+		"-files", extraBinMust("wpa_supplicant"),
+		"-files", extraBinMust("wpa_action"),
+		"-files", extraBinMust("wpa_cli"),
+		"-files", extraBinMust("wpa_passphrase"),
+	}
+	if *bzImage != "" {
+		args = append(args, "-files", *bzImage+":bzImage")
```
bjakobson: why path bzImage here, add`file := filepath.Join(user.home, "linux/arch/x86/boot/bzImage:bzImage)` or something that accomplishes that, then on line 44, make it this `"go", "run", "github.com/u-root/u-root/.", "file"`

rminnich: no, look at the args []string above. If you look at where it use to be you should be able to see I moved it, so I can add to it based on the bzImage and iso flags.

The args are used in the commands [][]string below. You can see how it used to be done. the commands array is what is used in the range commands for loop below. The args have to be the last thing run, after the go get commands.

Plus with what you are suggesting, we would not get all the other -files commands we want. 

Further, you only want to run u-root once

 ... and with your file := suggestion you would be forcing people to only be able to ever use $HOME/linux/arch/x86/boot/bzImage as their kernel, which is very inconvenient.

You should fetch this and try it out and you'll see how it works.

PR #53, opened 2020-01-27
Move the pmem to work on more systems

rminnich:
It's now at 1G, 512M in size. That's more than enough for most
.iso files, and few systems have less than 2G.

Linux from what we can see properly handles this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ qemu-system-x86_64 \
   -enable-kvm \
   -m 20G \
   -kernel linux/arch/x86/boot/bzImage \
-  -append 'console=ttyS0 console=tty1 memmap=4G!4G' \
+  -append 'console=ttyS0 console=tty1 memmap=1G!1G' \
   -initrd /tmp/initramfs.linux_amd64.cpio \
   -device virtio-rng-pci
 ```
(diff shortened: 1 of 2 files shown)


orangecms:
The Ubuntu ISO is actually quite huge, and when I tested even with 2GB for the memmap, it wasn't sufficient. I agree that half a GB would be enough for TinyCore, but that's really just because it's so tiny. :) Arch Linux ISOs are about 650M. Here's a list:

|distro|ISO size|
|----|----|
|Fedora | 1.8GB |
|openSUSE | 3.8GB |
|Ubuntu | 1.9GB |

orangecms:
Instead of the full desktop ISOs we could also consider netboot images, such as

- http://archive.ubuntu.com/ubuntu/dists/bionic-updates/main/installer-amd64/current/images/netboot/ `mini.iso` is 73MB
- https://software.opensuse.org/distributions/leap netboot is 125MB
- https://ftp.halifax.rwth-aachen.de/fedora/linux/releases/30/Workstation/x86_64/iso/ netinst is 600MB in size

orangecms:
Well, for openSUSE our purpose would be met with the [GNOME live image](https://download.opensuse.org/distribution/leap/15.1/live/openSUSE-Leap-15.1-GNOME-Live-x86_64-Current.iso), which is 931MB. So the 1GB would kinda work for that one, I suppose.

rminnich:
can we make it 1G!1G then?

PR #60, opened 2020-01-29
add run-webboot.sh script to launch QEMU for webboot

orangecms:


orangecms:
Why...?
```
2020/01/29 01:22:29 error building: error building go package in "/home/vsts/work/1/s/gopath/src/github.com/u-root/u-root/bb": # github.com/u-root/webboot/pkg/mountkexec
../../webboot/pkg/mountkexec/mountkexec.go:67:27: cannot use kernel (type *os.File) as type string in argument to multiboot.Probe
, exit status 2```

The code is untouched... :thinking:

bjakobson:
the reason is because I added kernel type *os.Open, because a change in u-root happened that caused multiboot.probe to not take in a string. it worked for me, even using the command, so not sure why it wouldn't wok for you

orangecms:
Yea until right now because u-root got a ton of reverts. -.-

bjakobson:
let me see something, ill check

orangecms:
We should fix this repo to not just YOLO use the master branch of u-root, however that works with Golang. You'll never know what you get otherwise.

bjakobson:
ok, it doesn't work if I don't do the change. Let me have you do something: run this: `go run buildimage.go` inside webboot, and if you get an error, let me know

bjakobson:
im curious about something, Azure is using go1.12, i wonder if this varies between versions

bjakobson:
i found the error, will fix in a sec

PR #62, opened 2020-01-29
Kexecmount revert

bjakobson:
U-root reverted the change they made to multiboot.go,causing building initramfs to be blocked due to `kernel` being type `*os.Open`, not `string`. Not a long-term fix, but needed until u-root commits again


bjakobson:
one second, this got reverted.

PR #64, opened 2020-01-31
build webboot standalone, not as part of u-root

rminnich:
The u-root churn has caused us a lot of trouble lately.
The problem is that we were building webboot as part of a u-root
build, and the fact that then u-root depended on webboot
which then depended on u-root was a headache.

We now build webboot as a standalone program, so it can use the
vendored u-root in this directory. This will end the mess we've had.

While we're at it, take away the auto-fetch of u-root and NiChrome.
It's a headache when offline.


rminnich:
I think this is the simplest way for now around the u-root churn: build webboot as an independent binary instead of as part of a u-root busybox. That allows u-root to change, while we stick with a vendor'ed snapshot. It's not quite working, but it should work ...

PR #67, opened 2020-02-02
Do some pre-checking first

rminnich:
It's annoying to wait for all kinds of things to happen only to notice
you typed a bad name, or /dev/pmem0 is not there.

Before we start messing with the network, at least make sure
the name is valid, and while we're at it, make sure there
is a /dev/pmem0.

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -34,12 +34,12 @@ import (
 )
 
 const (
-	wbtcpURL = "https://github.com/u-root/webboot-distro/raw/master/iso/tinycore/10.x/x86_64/release/TinyCorePure64.iso"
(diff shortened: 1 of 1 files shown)


bjakobson:
looks good! maybe this solved a weird thing i had: when launching from webboot into tinycore, it looks for eth0, and spends a stupid amount of time to give me an interface warning saying cannot find eth0, all interfaces configured. not really a “issue”, just pointing it out that it’s wasting time to know it doesn’t have eth0

rminnich:
that's a different problem I also want to fix :-)

rminnich:
I need to see more of that error, it would be useful, I don't get it

PR #70, opened 2020-02-02
Stop -cmd from overwriting tcCmdLine

bjakobson:
If a user needs to change any part of the booting process, they use the `-cmd` flag. This would actually overwrite the commands that came from tcCmdLine. 

For Example:

If you boot into tinycorepure using `webboot webboot-tinycorepure`, the command line that gets passed is this:

 `memmap=1G!1G earlyprintk=ttyS0,115200 console=ttyS0 console=tty0 root=/dev/pmem0 loglevel=3 cde`

 But say I want to change my memmap by altering -pmbase and -pmsize, so I boot by saying `webboot -cmd="-pmbase=4 -pmsize=4" webboot-tinycore`. But now, the command actually is only this:

`-pmbase=4 -pmsize=4`

this is now solved, so the command that get's passed is this:

-pmbase=4 -pmsize=4 memmap=4G!4G earlyprintk=ttyS0,115200 console=ttyS0 console=tty0 root=/dev/pmem0 loglevel=3 cde


webboot/webboot.go
```diff
@@ -207,6 +191,17 @@ func main() {
 		dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
 	}
 
+	arg := flag.Arg(0)
```
rminnich: you need to do a git pull --rebase

PR #72, opened 2020-02-02
use webboots cmdline as the base for all cmdline

bjakobson:
Almost everything is the same between `cat /proc/cmdline` in webboot's kernel and any other kernel. We can just use the outcome of `cat /proc/cmdline` as the basis of all other cmdlines.

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -34,16 +34,23 @@ import (
 )
 
 const (
-	wbtcpURL  = "https://github.com/u-root/webboot-distro/raw/master/iso/tinycore/10.x/x86_64/release/TinyCorePure64.iso"
-	wbcpURL   = "https://github.com/u-root/webboot-distro/raw/master/iso/tinycore/10.x/x86_64/release/CorePure64.iso"
-	tcURL     = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
-	coreURL   = "http://tinycorelinux.net/10.x/x86/release/CorePlus-current.iso"
-	ubuURL    = "http://releases.ubuntu.com/18.04/ubuntu-18.04.3-desktop-amd64.iso"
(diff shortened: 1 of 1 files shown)


webboot/webboot.go
```diff
@@ -34,16 +34,24 @@ import (
...
+	wbcpURL  = "https://github.com/u-root/webboot-distro/raw/master/iso/tinycore/10.x/x86_64/release/CorePure64.iso"
+	tcURL    = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
+	coreURL  = "http://tinycorelinux.net/10.x/x86/release/CorePlus-current.iso"
+	ubuURL   = "http://releases.ubuntu.com/18.04/ubuntu-18.04.3-desktop-amd64.iso"
+	archURL  = "http://mirror.rackspace.com/archlinux/iso/2020.01.01/archlinux-2020.01.01-x86_64.iso"
+
 )
 
 var (
+	commandline, err = ioutil.ReadFile("/proc/cmdline")
```
rminnich: you need to check the error, which means you can not do this in a declaration, but it is close. You can leave the DECLARATION of commandline here, and do the read in main()

webboot/webboot.go
```diff
@@ -34,16 +34,24 @@ import (
...
+
+	values = []string{string(commandline), " root=/dev/pmem0 vga=791"}
```
rminnich: do this in the tinyCoreCmdLine() func, not here.

webboot/webboot.go
```diff
@@ -84,25 +90,31 @@ var (
 		"arch": &webboot.Distro{
 			"arch/boot/x86_64/vmlinuz",
 			"/arch/boot/x86_64/archiso.img",
-			"memmap=1G!1G earlyprintk=ttyS0,115200 console=ttyS0 console=tty0 root=/dev/pmem0 loglevel=3",
+			*cdl,
```
rminnich: Just change the tinyCoreCmdLine func, for now, and leave this bookmark struct unchanged.

PR #73, opened 2020-02-06
added variable commandline to check /proc/cmdline

bjakobson:

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -178,6 +178,14 @@ func usage() {
 func main() {
 	flag.Parse()
 
+	cl, err := ioutil.ReadFile("/proc/cmdline")
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+
+
 	if flag.NArg() != 1 {
 		usage()
 	}
@@ -232,11 +240,9 @@ func main() {
 
 	}
 	if *dryrun == false {
-		if cmdline, err := webboot.CommandLine(bookmark[filename].Cmdline, *cmd); err != nil {
-			log.Fatalf("Error in webbootCommandline:%v", err)
-		} else {
-			bookmark[filename].Cmdline = cmdline
-		}
+	bookmark[filename].Cmdline = string(cl) + " " + bookmark[filename].Cmdline
+
+
 		if err := mountkexec.KexecISO(bookmark[filename], tmp); err != nil {
 			log.Fatalf("Error in kexecISO:%v", err)
 		}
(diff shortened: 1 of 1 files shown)


webboot/webboot.go
```diff
@@ -178,6 +178,19 @@ func usage() {
 func main() {
 	flag.Parse()
 
+	commandline, err := ioutil.ReadFile("/proc/cmdline")
+
+
+	if err != nil {
+		fmt.Errorf("Unable to read /proc/cmdline: %v", err)
+		//maybe here, we can have a hardcoded commandline so if
```
rminnich: nope. If we can't read /proc/cmdline we are done here.

webboot/webboot.go
```diff
@@ -178,6 +178,19 @@ func usage() {
 func main() {
 	flag.Parse()
 
+	commandline, err := ioutil.ReadFile("/proc/cmdline")
+
+
+	if err != nil {
+		fmt.Errorf("Unable to read /proc/cmdline: %v", err)
```
rminnich: This line has no effect. Just do this:
log.Fatal(err)
The error message will have that file name in it.
https://play.golang.org/p/xrfTNOGuzBM

webboot/webboot.go
rminnich: and now that you've see how this works, a slight set of changes. First, change this as follows
add an import
"github.com/u-root/u-root/pkg/cmdline"

Change this code to this:
		cl := cmdline.NewCmdLine()
		if cl.Err != nil {
			log.Fatal(cl.err)
		}

There is code at line 248:
change lines 248-252 as follows (yes, those 5 lines become 1 line)

			bookmark[filename].Cmdline = cl + " " + bookmark[filename].Cmdline

Later, we're going to fix the bookmarks to have minimal text, but this will do for now.

bjakobson: i don't know what you mean. you cannot take `cl` which is `cmdline.Cmdline` and put it into bookmark as a string

rminnich: yeah, sorry.
bookmark[filename].Cmdline = cl.Cmdline + " " + bookmark[filename].Cmdline

webboot/webboot.go
```diff
@@ -53,32 +53,30 @@ var (
...
 			"/boot/corepure64.gz",
-			tinyCoreCmdLine(),
+			tcCmdLine,
 			wbcpURL,
 		},
 		"tinycore": &webboot.Distro{
 			"boot/vmlinuz64",
 			"/boot/corepure64.gz",
-			tinyCoreCmdLine(),
+			tcCmdLine,
```
rminnich: are there trailing spaces/tabs here? be sure to run gofmt -w

PR #76, opened 2020-02-08
Remove functions we do not need

rminnich:
The various mount functions in mountkexec were doing work
we did not need. Remove them.

This is part of a larger project, to shrink the amount
of code in webboot we don't really need to write.


rminnich:
This is tested and working on qemu.

PR #78, opened 2020-02-09
add precheck function, check GO111MODULE

bjakobson:
With the `precheck` function implemented, we can check anything from the `go env`. 

Currently, we are using the `precheck` function to check `GO111MODULE`. If `GO111MODULE=on` or `GO111MODULE=off`, then `buildimage.go` will not work. Now, `precheck` will check `GO111MODULE`, and set it to `GO111MODULE=auto` if it's not already. 
Note: if `GO111MODULE=auto` or `GO111MODULE=""`, `precheck` won't do anything or change anything.

In the future, we can use `precheck` to check other things from the `go env`.


buildimage.go
```diff
@@ -36,6 +36,14 @@ var (
 	wifi    = flag.Bool("wifi", true, "include wifi tools")
 )
 
+//precheck can check anything from go env
```
rminnich: this is not checking anything. It's just a wrapper for os.Getenv

Why not just warn users but let them do what they want?

// preCheck checks the value of an environment variable against a set of allowed values and returns
// an error if none of them match.
func preCheck(val string, allowed ...string) error {
// you may or may not have seen this style of declaration. 
}

buildimage.go
```diff
@@ -54,6 +62,14 @@ func extraBinMust(n string) string {
 	return p
 }
 func main() {
+	//if GO111MODULE=off or GO111MODULE=on, buildimage won't work
+	//only GO111MODULE="" or GO111MODULE=auto works
+	s := precheck("GO111MODULE")
+
+	if s != "" || s != "auto" {
+		//set GO111MODULE=auto if it's not already, or GO111MODULE="".
+		os.Setenv("GO111MODULE", "auto")
```
rminnich: you should not os.Setenv, rather, add the correct value to the Env in commands you run.

buildimage.go
bjakobson: why “suggest”. there’s no point of suggesting to export it as off if it won’t work for them with it being off. or on. i think it should be set as auto or “”, or program exits

rminnich: Yes. Check it for a proper value and exit with a warning. if it is wrong.  I am pretty sure that's what I said.

Or see how the u-root command sets up CGO_ENABLED. You need not check; just set the proper value of GO111MODULE in the Env array of the exec.Cmd structs you use to run programs. If you are not familiar with how the Env array works in the Cmd struct it's time to take a look.

rminnich:
Also squash the commits!

PR #79, opened 2020-02-09
remove packages we created that we do not need

rminnich:


webboot/webboot.go
```diff
@@ -234,12 +243,33 @@ func main() {
...
-		if err := mountkexec.KexecISO(bookmark[filename], tmp); err != nil {
-			log.Fatalf("Error in kexecISO:%v", err)
+		kernel, err := os.Open(k)
+
+		if err != nil {
+			log.Fatal(err)
+		}
+		var image boot.OSImage
+		if err := multiboot.Probe(kernel); err == nil {
+			image = &boot.MultibootImage{
```
hugelgupf: This image can't be right

rminnich: It's in the original; shall I just yank it? What is not right.

rminnich: I mean the original webboot stuff. That's almost certainly wrong. I'm inclined to removed this until we care about multiboot.

webboot/webboot.go
```diff
@@ -234,12 +243,33 @@ func main() {
...
+		if err := image.Load(true); err != nil {
+			log.Fatalf("error failed to kexec into new kernel:%v", err)
+		}
+		if err := kexec.Reboot(); err != nil {
+			log.Fatalf("error failed to Reboot into new kernel:%v", err)
 		}
 	}
 
-	fmt.Printf("The URL requested: %v\n The file requested: %v\n The mounting point: %v\n", URL, filename, tmp)
+	fmt.Printf("URL: %v\n Distro: %v\nmount point: %v\n", b.URL, b, tmp)
```
bjakobson: why have each part be on separate lines? what’s wrong with having it all on one line?

webboot/webboot.go
rminnich: I remove dit ...

webboot/webboot.go
rminnich: it's showing one removed, one added; it's not two separate lines.

rminnich: oh I see what you mean now, oops. 
if you run it you can see. It's just more readable.

rminnich: ah, I see what you mean, now. 
We did it that way to make it more readable. If you try it you'll see.

PR #82, opened 2020-02-10
figuring out dhcp client

bjakobson:

### DIFF
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -212,7 +212,8 @@ func main() {
 			log.Fatal(err)
 		}
 	}
-
+	//the Request function sets up a DHCP confifuration for all interfaces,
+	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
 		dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
 	}
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -215,7 +215,8 @@ func main() {
 	//the Request function sets up a DHCP confifuration for all interfaces,
 	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
-		//ifname uses the regular expression ^[we].* to check for wlan0 or eth0 interfaces.
+		//ifname uses the regular expression ^[we].* to check for an interface starting with w or e such as
+		//wlan0 or eth0
 		dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
 
 	
--- a/webboot/webboot.go
+++ b/webboot/webboot.go
@@ -215,11 +215,7 @@ func main() {
 	//the Request function sets up a DHCP confifuration for all interfaces,
 	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
-		//ifname uses regex to find patterns.
-		//it uses ^ as the start of the line,
-		//[we] to find any single character containing w,e
-		//. to find any single charachter
-		//and * suggesting 0 or more or ^[we].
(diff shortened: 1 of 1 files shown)


webboot/webboot.go
```diff
@@ -212,7 +212,8 @@ func main() {
 			log.Fatal(err)
 		}
 	}
-
+	//the Request function sets up a DHCP confifuration for all interfaces,
+	//such as eth0, which is an ethernet interface.
```
rminnich: now for each argument, go ahead and describe in your own words what it's for. I know that may seem dumb but bear with me here. 
e.g. 
*ifName is the name of the interface
[hint: that's wrong. Look carefully at the default value of the argument to see why]

webboot/webboot.go
```diff
@@ -212,9 +212,12 @@ func main() {
...
 	if *ipv4 || *ipv6 {
 		dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
+
+		//*ifname is used as the name of the interface, defaulting with ^[we].*
```
rminnich: if you can, put the comment above the call i.e. before line 218. And, note, ... "name" is not right. It's singular. So now you need to dig a bit and see ... could it be more than one? If so, why?

bjakobson: 👍

webboot/webboot.go
rminnich: but how? And it's not for all interfaces. Why is the name so weird?
^[we].*

and how does that end up being for all interfaces? Not to be mean but I suspect there's a bit of guessing in your answer. Not surprising or bad if so; this is a SIMPLE thing that hides a COMPLEX answer.

webboot/webboot.go
```diff
@@ -212,9 +212,17 @@ func main() {
...
 	}
-
+	//the Request function sets up a DHCP confifuration for all interfaces,
+	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
+		//ifname uses regex to find patterns.
+		//it uses ^ as the start of the line,
+		//[we] to find any single character containing w,e
+		//. to find any single charachter
+		//and * suggesting 0 or more or ^[we].
```
rminnich: almost!!!
^ - start of line
[we] --   either one
. -- any character
* -- modifies the ., not the [we]

here's how to play with regex
grep '^[we].*'
then start typing typical names and see what happens
e.g.
grep '^[we].*'
eth0
wlan09
alan0

and so on

you're close

bjakobson: i haven't tried it (not home).... but i'm guessing the `w` in `^[we].*` is for wlan0 and `e` is eth0... just a hunch.

rminnich: which is true, but your description of the * is wrong. the . and * and what they mean are supremely important ...

bjakobson: . is any character 
* is expect more than ^[we] 

(this is what https://yourbasic.org/golang/regexp-cheat-sheet/ says)

webboot/webboot.go
```diff
@@ -212,9 +212,13 @@ func main() {
 			log.Fatal(err)
 		}
 	}
-
+	//the Request function sets up a DHCP confifuration for all interfaces,
+	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
+		//ifname uses the regular expression ^[we].* to check for wlan0 or eth0 interfaces.
```
rminnich: that re matches more than just wlan0 and eth0 right?

bjakobson: well i don’t know what exactly w and e can stand for, but i know wlan and eth0 are options. maybe those are just examples but i don’t know what else could be.

PR #84, opened 2020-03-29
Delete webboots pkg/dhclient to use u-root's pkg/dhclient

bjakobson:
modified webboot/webboot.go in order to use u-root's dhclient. works to build initramfs.


webboot/webboot.go
```diff
@@ -212,12 +219,47 @@ func main() {
...
+	if *verbose {
+		c.LogLevel = dhclient.LogSummary
+	}
+	
 	//the Request function sets up a DHCP confifuration for all interfaces,
 	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
 		//ifname uses the regular expression ^[we].* to check for an interface starting with w or e such as
 		//wlan0/1, enx453243, or eth0/1
+<<<<<<< Updated upstream
```
rminnich: you need to squash the commits into one. You've got a problem right here with this <<< and such. It all seems to be fixed in the final commit but this stuff should never be in a commit.

It works fine for me, on qemu, so well done!

PR #85, opened 2020-04-03
Delete webboots pkg/dhclient to use u-root's pkg/dhclient

bjakobson:
modified webboot/webboot.go in order to use u-root's dhclient. works to build initramfs.


bjakobson:
> and chance you can git rebase -i origin/master and squash this into one commit?

no. i messed up on these commits, but I know for next time what went wrong

rminnich:
ah, what I mean is
you're going to have to git rebase -i origin/master and learn how to squash commits, it can't go in this way. And it's easy. Worth learning to do.

bjakobson:
i know how to squash commits via rebase. my issue is this error: ‘ error: cannot 'squash' without a previous commit’.

webboot/webboot.go
```diff
@@ -212,13 +220,50 @@ func main() {
...
+
 	//the Request function sets up a DHCP confifuration for all interfaces,
 	//such as eth0, which is an ethernet interface.
 	if *ipv4 || *ipv6 {
 		//ifname uses the regular expression ^[we].* to check for an interface starting with w or e such as
 		//wlan0/1, enx453243, or eth0/1
-		dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
 
+		//dhclient.Request(*ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
+		dhclient.IfUp(*ifname)
```
rminnich, line 265: this is wrong. IfUp takes a name of an interface and you are passing it the regular expression. It's also not needed here, so not sure why you added it.

webboot/webboot.go
```diff
@@ -212,13 +220,50 @@ func main() {
...
+		dhclient.SendRequests(ctx, filteredIfs, *ipv4, *ipv6, c)
```
rminnich, line 266: you are ignoring the return value of sendrequests. You should wait for at least one successful return value, so that you know at least one interface is up and ready. Failure to do this is what panics my kernels.

webboot/webboot.go
```diff
@@ -57,7 +65,7 @@ type Distro struct {
 
 var (
 	cmd      = flag.String("cmd", "", "Command line parameters to the second kernel")
-	ifName   = flag.String("interface", "^[we].*", "Name of the interface")
+	ifname   = flag.String("interface", "^[we].*", "Name of the interface")
```
rminnich, line 68: it's not the name of the interface, it's an RE describing one or more interfaces.

webboot/webboot.go
```diff
@@ -28,13 +28,21 @@ import (
...
+
 
 	"github.com/u-root/u-root/pkg/boot"
 	"github.com/u-root/u-root/pkg/boot/kexec"
 	"github.com/u-root/u-root/pkg/mount"
 	"github.com/u-root/u-root/pkg/uio"
-	"github.com/u-root/webboot/pkg/dhclient"
+	"github.com/u-root/u-root/pkg/dhclient"
 	"golang.org/x/sys/unix"
+	"github.com/vishvananda/netlink"
```
rminnich, line 42: you need to work to change this code so that we do not need to import netlink.

rminnich:
some very good news. If you can look at my comment re line 266, where I mention you need to wait to get at least one value back from the dhcp, ... this fixes my problems and I just booted to graphics on my thinkpad t510

webboot/webboot.go
bjakobson: webboot has been using that description the whole time. I can change it, I really just changed ifName to ifname and left the values the same.

webboot/webboot.go
bjakobson: will do.

webboot/webboot.go
rminnich: yes, my point here is you were using this re below in a function that took a name.

Change this variable name to ifRE meaning interface Regular Expression.

bjakobson:
i'll look, but i don't see why. I am using the same exact functions that webboot/dhclient.go used except instead of using in functions, im hard coding the values.. *ifRe, *timeout instead of ifRe or timeout. webboot/dhclient.go imported netlink.

PR #93, opened 2020-06-16
add cmds/webboot

xsy1996:
build a menu of iso and display the path of iso's kernel.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -16,33 +16,38 @@ import (
 	"github.com/u-root/webboot/pkg/menus"
 )
 
-// kernelPath is for hardcode the path to kernel. should be replaced by u-root later
-var kernelPath = make(map[string]string)
-
-
 const (
-	tcURL     = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
+	tcURL = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
 )
 
-type Distro struct {
-	URL     string
+//DownloadDistro contains information of downloadable isos.
+type DownloadDistro struct {
+	URL string
 }
 
 // bookmark record the list of iso webboot allow user to download,
 // and users can enter these label to indicate an iso to download.
-var bookmark = map[string]*Distro{
-	"tinycore": &Distro{
+var bookmark = map[string]*DownloadDistro{
+	"tinycore": &DownloadDistro{
 		tcURL,
-	}
+	},
+}
+
+// kernelPath is for hardcode the path to kernel. should be replaced by u-root later
+var kernelPath = map[string]string{
+	"archlinux-2020.06.01-x86_64.iso": "arch/boot/x86_64/vmlinuz",
+	"TinyCore-11.1.iso":               "boot/vmlinuz",
 }
 
 var (
 	isoDir  = flag.String("dir", "", "set the iso directory path")
 	timeout = flag.Int("timeout", 15, "Lease timeout in seconds")
 	retry   = flag.Int("retry", 5, "Max number of attempts for DHCP clients to send requests. -1 means infinity")
-	verbose = flag.Bool("verbose", false, "Verbose output")
+	v       = flag.Bool("verbose", false, "Verbose output")
 	ipv4    = flag.Bool("ipv4", true, "use IPV4")
 	ipv6    = flag.Bool("ipv6", true, "use IPV6")
+
+	verbose = func(string, ...interface{}) {}
 )
 
 // ISO type contains information of a iso file, incluing its name,
@@ -61,6 +66,8 @@ type DownloadOption struct {
 	url     string
 }
 
+// DownloadByLinkOption contains information of the iso user want to download,
+// include it's name and the download URL
 type DownloadByLinkOption struct {
 	isoName string
 	url     string
@@ -81,13 +88,13 @@ func (u ISO) Label() string {
 }
 
 // Do is what will be called after user choose this iso option,
-func (u ISO) Do() error{
+func (u ISO) Do() error {
 	if err := getKernel(&u); err != nil {
 		return err
 	}
-	if(*verbose){
-		log.Printf("isoPath is %s, isoKernel is %s", chosenISO.path, chosenISO.kernel)
-	}
+	verbose("isoPath is %s, isoKernel is %s", u.path, u.kernel)
+
+	return nil
 }
 
 // IsDefault is for mark whether this entry is a default choice.
@@ -107,144 +114,132 @@ func (u DownloadOption) Label() string {
 }
 
 // Do is what will be called after user choose this download option,
-func (u DownloadOption) Do() error{
+func (u DownloadOption) Do() error {
 	// If download option is chosen, user can input the name of iso they want to download.
 	// The name of iso should be include in the bookmark.
-	isoname_check_func = func(input string) (bool, string) {
-		if _, ok := downloadList[input]; ok {
+	isonameCheckFunc := func(input string) (bool, string) {
+		if _, ok := bookmark[input]; ok {
 			return true, ""
 		}
 		return false, "not able to find this distro. please try another."
 	}
-	isoLabel, err := menus.NewInputWindow("Please input the name of iso you want to download and install here:", 100, 1, isoname_check_func)
+	isoLabel, err := menus.NewInputWindow("Please input the name of iso you want to download and install here:", 100, 1, isonameCheckFunc)
 	if err != nil {
-		log.Println(err)
-		return "", ""
-	}
-	downloadOption.isoName = isoLabel
-	downloadOption.url = downloadList[isoLabel].URL
-
-	if(*verbose){
-		log.Printf("arrave download step, URL of iso is %s\n", downloadOption.url)
-	}
-	if err := downloadIso(filepath.Join(*isoDir, isoLabel+".iso"), downloadOption.url); err!=nil{
 		return err
-	}else if(*verbose){
-		log.Printf("iso is downloaded")
 	}
+	u.isoName = isoLabel
+	u.url = bookmark[isoLabel].URL
+
+	verbose("arrave download step, URL of iso is %s\n", u.url)
+
+	if err := downloadIso(filepath.Join("/tmp/", u.isoName+".iso"), u.url); err != nil {
+		return err
+	}
+	verbose("iso is downloaded")
+	return nil
 }
 
+// IsDefault is for mark whether this entry is a default choice.
 func (u DownloadByLinkOption) IsDefault() bool {
 	return true
 }
 
-func (u DownloadODownloadByLinkOptionption) String() string {
+func (u DownloadByLinkOption) String() string {
 	return fmt.Sprintf("%+v\n", u)
 }
 
+// Label is the string this entry displays in the menu page.
 func (u DownloadByLinkOption) Label() string {
 	return "Download by link"
 }
 
-func (u DownloadOption) Do() error{
-	// if download by link option is chosen, user can input a link to download the iso. 
-	string_check_func = func(input string) (bool, string) {
+// Do is what will be called after user choose this download option,
+func (u DownloadByLinkOption) Do() error {
+	// if download by link option is chosen, user can input a link to download the iso.
+	stringCheckFunc := func(input string) (bool, string) {
 		return true, ""
 	}
-	isoLink, err := menus.NewInputWindow("Please input the link of iso you want to download and install here:", 100, 1, string_check_func)
+	isoLink, err := menus.NewInputWindow("Please input the link of iso you want to download and install here:", 100, 1, stringCheckFunc)
 	if err != nil {
 		return err
 	}
-	isoName, err := menus.GetInput("Please input the name of iso:", 4, 100, 1, string_check_func)
+	isoName, err := menus.NewInputWindow("Please input the name of iso:", 100, 1, string_check_func)
 	if err != nil {
 		return err
 	}
-	downloadByLinkOption.isoName = isoName
-	downloadByLinkOption.url = isoLink
-	if(*verbose){
-		log.Printf("arrave download step, URL of iso is %s\n", downloadByLinkOption.url)
-	}
-	if err := downloadIso(filepath.Join(*isoDir, isoName+".iso"), downloadByLinkOption.url); err!=nil{
+	u.isoName = isoName
+	u.url = isoLink
+	verbose("arrave download step, URL of iso is %s\n", u.url)
+	if err := downloadIso(filepath.Join("/tmp/", u.isoName+".iso"), u.url); err != nil {
 		return err
-	}else if(*verbose){
-		log.Printf("iso is downloaded")
 	}
-
+	verbose("iso is downloaded")
+	
+	return nil
 }
 
 // getKernel is to find the kernel of a ISO. if the path of ISO is given,
 // just check if the kernel is exist.
 func getKernel(u *ISO) error {
 
-	if(*verbose){
-		log.Println("try mount iso...")
-	}
+	verbose("try mount iso...")
 
 	isoName := u.name
 
 	diskFile, err := ioutil.TempDir("", "/mnt-iso")
 	if err != nil {
-		return   err
+		return err
 	}
 	defer os.RemoveAll(diskFile)
 
 	mountPath := filepath.Join(diskFile, isoName)
 	if mp, err := mount.Mount(u.path, mountPath, "iso9660", "", mount.ReadOnly); err != nil {
-		return fmt.Errorf(("TryMount %s = %v, want nil", u.path, err)
-	} else {
-		if(*verbose){
-			log.Printf("mounted disk file is %s\n", mountPath)
-		}
-		// if kernel is given, check its validation
-		if u.kernel != "" {
-			walkfunc := func(path string, info os.FileInfo, err error) error {
-				if path == filepath.Join(mountPath, u.kernel) {
-					log.Printf("Kernel at %s is found\n", u.kernel)
-				}
-				return nil
+		return fmt.Errorf("TryMount %s = %v, want nil", u.path, err)
+	}
+	verbose("mounted disk file is %s\n", mountPath)
+	// if kernel is given, check its validation
+	if u.kernel != "" {
+		walkfunc := func(path string, info os.FileInfo, err error) error {
+			if path == filepath.Join(mountPath, u.kernel) {
+				log.Printf("Kernel at %s is found\n", u.kernel)
 			}
-
-			if(*verbose){
-				log.Println("\nfinding kernel...")
-			}
-			filepath.Walk(diskFile, walkfunc)
-			if(*verbose){
-				log.Printf("Kernel path is %s\n", u.kernel)
-			}
+			return nil
 		}
 
-		log.Println("\ntry unmount iso...")
-		if err := mp.Unmount(0); err != nil {
-			return fmt.Errorf("Unmount(%q) = %v, want nil", mountPath, err)
-		}
+		verbose("\nfinding kernel...")
+		filepath.Walk(diskFile, walkfunc)
+		verbose("Kernel path is %s\n", u.kernel)
 	}
 
-	if(*verbose){
-		log.Println("Done")
+	log.Println("\ntry unmount iso...")
+	if err := mp.Unmount(0); err != nil {
+		return fmt.Errorf("Unmount(%q) = %v, want nil", mountPath, err)
 	}
+
+	verbose("Done")
 	return nil
 }
 
 // isoMenu is to  build the iso menu page and return the iso path
 // and kernel path of which user choose.
-func isoMenu(isos []ISO) (error) {
+func isoMenu(isos []ISO) error {
 
 	var entries []menus.Entry
 	for _, iso := range isos {
 		entries = append(entries, iso)
 	}
-	var downloadOption DownloadOption{"Download"}
-	var downloadByLinkOption DownloadOption{"Download By Link"}
+	var downloadOption DownloadOption
+	var downloadByLinkOption DownloadByLinkOption
 	entries = append(entries, downloadOption, downloadByLinkOption)
 
-	index, err := menus.DisplayMenu("ISO Menu", "Choose an iso you want to boot (hit enter to choose the default - 1 is the default option) >", 0, entries)
+	entry, err := menus.DisplayMenu("ISO Menu", "Choose an iso you want to boot (hit enter to choose the default - 1 is the default option) >", 0, entries)
 
 	// if index is exceed the length of entries or less than 0 means something wrong happened
 	if err != nil {
 		return err
-	}else if index < 0 || index >= len(entries) {
-		return fmt.Errorf("input index is less than 0 or exceed the length of option list")
 	}
+	err = entry.Do()
+	return err
 	return nil
 }
 
@@ -280,7 +275,7 @@ func downloadIso(isoPath, URL string) error {
 		return err
 	}
 	if *ipv4 || *ipv6 {
-		dhclient.Request(ifName, *timeout, *retry, *verbose, *ipv4, *ipv6)
+		dhclient.Request(ifName, *timeout, *retry, *v, *ipv4, *ipv6)
 	}
 	isoReader, err := linkOpen(URL)
 	if err != nil {
@@ -290,33 +285,30 @@ func downloadIso(isoPath, URL string) error {
 	isofile, err := os.Create(isoPath)
 	if err != nil {
 		return err
-	} else {
-		if _, err := io.Copy(isofile, isoReader); err != nil {
-			return fmt.Errorf("Error copying to persistent memory device: %v", err)
-		}
-		if err = isofile.Close(); err != nil {
-			return fmt.Errorf("Error closing %s: %v", isoPath, err)
-		}
+	}
+	if _, err := io.Copy(isofile, isoReader); err != nil {
+		return fmt.Errorf("Error copying to persistent memory device: %v", err)
+	}
+	if err = isofile.Close(); err != nil {
+		return fmt.Errorf("Error closing %s: %v", isoPath, err)
 	}
 
-	if(*verbose){
-		log.Printf("downloaded isoPath is %s", isoPath)
-	}
+	verbose("downloaded isoPath is %s", isoPath)
 	return nil
 }
 
 func main() {
 	flag.Parse()
 
-	// hardcode kernel path
-	kernelPath["archlinux-2020.06.01-x86_64.iso"] = "arch/boot/x86_64/vmlinuz"
-	kernelPath["TinyCore-11.1.iso"] = "boot/vmlinuz"
+	if *v {
+		verbose = log.Printf
+	}
 
 	var isos []ISO
 
-	// If dir is null, that means we need use findDevice to find the USB disk
+	// If isoDir is not given, that means we need use findDevice to find the USB disk
 	// else we walk through the given folder to find all iso inside.
-	if(*dir == ""){
+	if *isoDir == "" {
 
 	} else {
 		walkfunc := func(path string, info os.FileInfo, err error) error {
@@ -330,8 +322,8 @@ func main() {
 	}
 
 	err := isoMenu(isos)
-	if(err != nil){
-		log.Fatalf(err)
+	if err != nil {
+		log.Fatalf("%v", err)
 	}
 	return
 }
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -1,0 +1,245 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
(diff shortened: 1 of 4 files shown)


cmds/webboot/ISO-menu.go
```diff
@@ -0,0 +1,269 @@
...
+
+//the entry interface record all information of a menu choise
+type Entry interface {
+	Label() string // the label will show in menu
+	Do() error  //do func will be run when the entry is chosen
+	IsDefault() bool  // IsDefault is true means this entry will be hit by default, if there is many default choise, the first on in the list will be choose. the rest will be ignored.
+}
+
+//GetChoose will present all entries into a menu with numbers, so that user can choose from them.
+func GetChoose(coord *int, introwords string, entries ...Entry) error{
```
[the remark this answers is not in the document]
pallaud: And actually related to that, I think it might make sense to separate the interface/menu elements from the rest of the functionality altogether. What I mean by that is instead of having `GetChoose` call a Do() method for your entry, instead have it only take a list of labels to display, and return the label that was selected. Then from here, you can mount the one that was chosen. Does that make sense? That way, all the the menu function is doing is displaying options in a menu and returning the one that was chosen. I think that would mean you don't need to have the Entry type anymore either.

cmds/webboot/ISO-menu.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (u ISO) String() string {
+    return fmt.Sprintf("%s, %s, %s, %v\n", u.name, u.path, u.kernel, u.isDefault)
+}
+
+func (u ISO) Label() string {
+    return u.name
+}
+
+var (
```
pallaud, line 219: You usually want to declare global variables and types at the top of the file!

cmds/webboot/ISO-menu.go
```diff
@@ -0,0 +1,269 @@
...
+)
+
+
+func main() {
+
+
+    flag.Parse()
+
+    //kernelPath is for hardcode the path to kernel. should be replaced by u-root later
+    kernelPath := make(map[string]string)
```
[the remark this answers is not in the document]
xsy1996: what is the benefit to make kernelPath a global variable?

cmds/webboot/ISO-menu.go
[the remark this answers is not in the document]
xsy1996: the entry interface is imitate the menu of u-root. I don't know why u-root will use this entry interface instead of a single string(I think string is much more easy too) but I guess maybe  the goal of having a Do() func is for concurrency or something?

cmds/webboot/ISO-menu.go
```diff
@@ -0,0 +1,269 @@
...
+	ui "github.com/gizak/termui/v3"
+	"github.com/gizak/termui/v3/widgets"
+)
+
+
+//the entry interface record all information of a menu choise
+type Entry interface {
+	Label() string // the label will show in menu
+	Do() error  //do func will be run when the entry is chosen
+	IsDefault() bool  // IsDefault is true means this entry will be hit by default, if there is many default choise, the first on in the list will be choose. the rest will be ignored.
```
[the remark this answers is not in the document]
xsy1996: sure I can do this, but I just wonder will we want to indicate 1 or more specific default iso in the future?

cmds/webboot/ISO-menu.go
```diff
@@ -0,0 +1,269 @@
...
+    if  err != nil {
+        // ui.Close()
+        fmt.Println(err)
+        // os.Exit(1)
+    }
+    diskName := strings.Fields(string(output))[0]
+    diskFile := "/mnt/iso"
+    fmt.Println("iso filename is "+u.path)
+    fmt.Println("mounted disk name is "+diskName)
+    cmd = exec.Command("mount","-t","cd9660",diskName,diskFile)
```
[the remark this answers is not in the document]
xsy1996: thanks! that's useful! I'll try this package

cmds/webboot/ISO-menu.go
```diff
@@ -0,0 +1,269 @@
...
+    isDefault bool
+}
+
+func (u ISO) IsDefault() bool {
+    return u.isDefault
+}
+
+//Do func is what to do when an iso is chosen.
+func (u ISO) Do() error{// these command below is only work for OSX now. I plan to add a linux version and decided which to use by GOOS. hope it will work.  
+    ui.Close() // close the old window and open a new one
```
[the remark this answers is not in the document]
xsy1996: emm I am not understand what you mean, maybe we can discuss it in a chat?

pkg/menus/menus.go
```diff
@@ -0,0 +1,146 @@
...
+    }
+    listData := []string{}//listData contains all choice's labels
+    
+    for i, e := range entries {
+        listData = append(listData, fmt.Sprintf("[%d] %s", i+1, e.Label()))
+	}
+
+    //l is a List item which will present like a menu by termui package
+    l := widgets.NewList()	
+    l.Title = "Menu"
```
pallaud, line 35: Maybe the title  is also something you could take as an input?

pkg/menus/menus.go
```diff
@@ -0,0 +1,146 @@
...
+	                }
+
+    				ui.Close()
+	                //if no one is default, choose the first one. 
+                    return 0, nil 
+                }
+                // try convert the input into a vilid number
+                c, err := strconv.Atoi(p1.Text) 
+                if(err!=nil || c<1 || c>len(entries)){
+                    p1.Text = fmt.Sprintf("%s is not a valid entry number.", p1.Text)
```
pallaud, line 89: I might not be understanding how termui works, but why do we need to use the same paragraph for user input and for displaying errors? Would it make sense to have a separate paragraph widget for that, and then on Enter, if the user input is invalid, call a separate function to display the error text, and reset the user input paragraph to ""?

pallaud: I think that way you could use your separate GetInput function, maybe makes this cleaner

cmds/webboot/iso-menu.go
```diff
@@ -0,0 +1,138 @@
...
+func isoMenu(isos []ISO) (string, string) {
+
+        coord := 0
+
+        var entries []menus.Entry
+        for _, iso := range isos {
+                entries = append(entries, iso)
+        }
+
+        index, err := menus.GetChoose(&coord, "Choose an iso you want to boot (hit enter to choose the default - 1 is the default option) >", entries)
```
pallaud, line 96: Why do we need to pass coord here?

xsy1996: These is because i need coord to tell me what is the loaction of the previous menu. but since we only have one menu per page it is useless now. I will remove it, sorry.

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+		c, err := strconv.Atoi(choose)
+		ui.Close()
+		return c - 1, err
+	}
+	ui.Close()
+	return -1, nil
+}
+
+// GetInput will present an input box to user and return the user's input.GetInout will check validation of input using checkValidFunc.
+func GetInput(introwords string, location int, inputWidth int, inputHeight int, checkValidFunc func(string) (bool, string)) (string, error) {
```
pallaud, line 89: Nit: In Go the convention tends to be to have shorter names, so maybe something like 'desc' instead of introwords and 'isValid' instead of checkValidFunc would be good?

pallaud: In general nice to have shorter variable names!

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+	Label() string
+	// IsDefault is true means this entry will be hit by default, if there is many default choise, the first on in the list will be choose. the rest will be ignored.
+	IsDefault() bool
+}
+
+// DisplayMenu will present all entries into a menu with numbers, so that user can choose from them.
+func DisplayMenu(menuTitle string, introwords string, location int, entries []Entry) (int, error) {
+	// open a new window
+	if err := ui.Init(); err != nil {
+		log.Printf("failed to initialize termui: %v", err)
```
pallaud, line 27: Again why print the error instead of returning it?

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// checkValidFunc is a function for GetInput to check the format of input. Since we want user to choose a option from the menu, the checkValidFunction will check :1.input is a number; 2.input number does not exceed the number of options.
+	checkValidFunc := func(input string) (bool, string) {
+		if input == "" {
+			for _, en := range entries {
+				if en.IsDefault() {
+					return true, ""
+				}
+			}
+			return false, "no default option, please enter a choice"
```
pallaud, line 54: I actually don't think we need this, I think we can always default to the first entry, that gets rid of the need for logic if there isn't a choice made - what do you think?

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+			return false, fmt.Sprintf("your input is not a valid entry number.")
+		}
+        return true, ""
+	}
+
+	// call GetInput to get user's choice
+	choose, err := GetInput(introwords, location, 100, 1, checkValidFunc)
+
+	if err != nil {
+		log.Printf("%v", err)
```
pallaud, line 67: Why do we need to print the error here if we're returning it anyway?

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+func DisplayMenu(menuTitle string, introwords string, location int, entries []Entry) (int, error) {
+	// open a new window
+	if err := ui.Init(); err != nil {
+		log.Printf("failed to initialize termui: %v", err)
+	}
+	// listData contains all choice's labels
+	listData := []string{}
+
+	for i, e := range entries {
+		listData = append(listData, fmt.Sprintf("[%d] %s", i+1, e.Label()))
```
pallaud, line 33: You could also start the menu numbering from 0 if that makes life easier :)

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	var downloadOption DownloadOption
+	entries = append(entries, downloadOption)
+
+	index, err := menus.DisplayMenu("ISO Menu", "Choose an iso you want to boot (hit enter to choose the default - 1 is the default option) >", 0, entries)
+
+	// if index is exceed the length of entries or less than 0 means something wrong happened
+	if err != nil || index < 0 || index >= len(entries) {
+		if err != nil {
+			log.Println(err)
```
pallaud, line 128: Why not return the errors from this function and then log.Fatalf in main instead of just printing here?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+	for _, iso := range isos {
+		entries = append(entries, iso)
+	}
+	var downloadOption DownloadOption
+	entries = append(entries, downloadOption)
+
+	index, err := menus.DisplayMenu("ISO Menu", "Choose an iso you want to boot (hit enter to choose the default - 1 is the default option) >", 0, entries)
+
+	// if index is exceed the length of entries or less than 0 means something wrong happened
+	if err != nil || index < 0 || index >= len(entries) {
```
pallaud, line 126: Nit: I think itd actually be cleaner if you split this into 2 if statements

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+// Entry contains all the information needed for a boot entry
+type Entry interface {
+	// Label is the string will show in menu
+	Label() string
+	// IsDefault is true means this entry will be hit by default, if there is many default choise, the first on in the list will be choose. the rest will be ignored.
+	IsDefault() bool
+}
+
+// DisplayMenu will present all entries into a menu with numbers, so that user can choose from them.
+func DisplayMenu(menuTitle string, introwords string, location int, entries []Entry) (int, error) {
```
pallaud, line 24: Instead of returning a number here, why not return the chosen entry? I'm not sure if it'd be better, just wondering

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+		if err != nil {
+			log.Println(err)
+		}
+		return "", ""
+	}
+	// if index is the same as the length of entries - 1, it means the chosen option is download. Elsewise the chosen option is a cached iso.
+	if index == len(entries)-1 {
```
pallaud, line 133: I commented this somewhere else too, but if DisplayMenu returned the entry chosen instead of just the index, this logic could be cleaner

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+		// if download option is chosen, webboot will show another input window for user to indicate what they want
+		isoLabel, err := menus.NewInputWindow("Please input the name of iso you want to download and install here:", 100, 1, func(input string) (bool, string) {
```
pallaud, line 135: Nit: Could you declare the func here as a separate variable and then pass that the menus.NewInputWindow so that this is a little easier to read?

pkg/menus/menus.go
xsy1996: because the entry do not contain all information I want, for example , a entry do not contain the path to the chosen iso. so I need the index  instead of entry

cmds/webboot/webboot.go
xsy1996: I hope I can return the entry, but entry do not contain enough information. And because different option need different information (like a iso option need the path of iso and the download option need the input name of iso) i can not make a entry work for all options. that is entry only contains option's label and whether it is a default option, but I need more information in most of situations, that's why I return index instead of entry.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/u-root/u-root/pkg/mount"
+	"github.com/u-root/webboot/pkg/dhclient"
+	"github.com/u-root/webboot/pkg/menus"
+)
+
+// kernelPath is for hardcode the path to kernel. should be replaced by u-root later
+var kernelPath = make(map[string]string)
+
+// downloadList record the list of iso webboot allow user to download, and users can enter these label to indicate an iso to download. I know that may be not flexiable enough, but I can not come up with a better way to let webboot know where it can download a distro without this list. Allow user directly input the website to download iso? i don't think that's a good idea too.
+var downloadList = map[string]string{"core": "http://tinycorelinux.net/10.x/x86/release/CorePlus-current.iso"}
```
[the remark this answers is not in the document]
rminnich: I agree. Also, this comment on line 22 needs to be fixed. Line 22 is 379 chars long and that is too long.

Users should be able to type an expression like http://tinycorelinux.net/10.x/x86/release/CorePlus-current.iso as an argument to webboot and have it load it.

Also, you should look at the bookmarks map in the original webboot, it has this map already.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+
+// ISO type contains information of a iso file, incluing its name, its filepath, the path to its kernel and if it should be choose by default.
+type ISO struct {
+	name      string
+	path      string
+	kernel    string
+	isDefault bool
+}
+
+//DownloadOption contains information of the iso user want to download, include it's name and the download URL
```
rminnich, line 37: All your comments will need to be fixed, else people will complain.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+type DownloadOption struct {
+	isoName string
+	url     string
+}
+
+// IsDefault is for mark whether this iso is a default choise.
```
rminnich, line 43: choice, not choise :-)

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+		// if kernel is given, check its validation
+		if u.kernel != "" {
+			walkfunc := func(path string, info os.FileInfo, err error) error {
+				if path == filepath.Join(mountPath, u.kernel) {
+					log.Printf("Kernel at %s is found\n", u.kernel)
+				}
+				return nil
+			}
+
+			log.Println("\nfinding kernel...")
```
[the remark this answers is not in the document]
rminnich: I think you should have a verbose option, but in error cases, you should return an error, not print.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		return resp.Body, nil
+	}
+	return nil, fmt.Errorf("%q: linkopen only supports file://, https://, and http:// schemes", URL)
+}
+
+// downloadIso will download a iso from URL and save it under savePath directory
+func downloadIso(isoPath, URL string) error {
+	log.Printf("Should begin to download here..., download url is %s\n", URL)
+	timeout := 15
```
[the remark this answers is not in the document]
rminnich: they could be flags, even.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+	return nil
+}
+
+func main() {
+	flag.Parse()
+
+	// hardcode kernel path
+	kernelPath["archlinux-2020.06.01-x86_64.iso"] = "arch/boot/x86_64/vmlinuz"
```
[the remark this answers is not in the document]
rminnich: yes, please see the bookmarks in the original webboot.

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+	"fmt"
+	"log"
+	"os"
+	"strconv"
+
+	ui "github.com/gizak/termui/v3"
+	"github.com/gizak/termui/v3/widgets"
+)
+
+var menuWidth = 50
```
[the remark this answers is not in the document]
rminnich: Or, better, find the width of the terminal you are on and use that. But a const is fine for now.

pkg/menus/menus.go
rminnich: I would argue you need to put more information in the entry, if it does not have enough information.

pkg/menus/menus.go
rminnich: yep.

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+	l := widgets.NewList()
+	l.Title = menuTitle
+	l.Rows = listData
+	// design the size and location of the menu
+	l.SetRect(0, location, menuWidth, location+len(entries)+2)
+	location += len(entries) + 2
+	l.TextStyle.Fg = ui.ColorWhite
+	ui.Render(l)
+
+	// checkValidFunc is a function for GetInput to check the format of input. Since we want user to choose a option from the menu, the checkValidFunction will check :1.input is a number; 2.input number does not exceed the number of options.
```
rminnich, line 46: This comment needs to be fixed. I wonder what editor you are using?

pkg/menus/menus.go
rminnich: inputWidth -> wid
inputHeight -> ht
and so on

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+	uiEvents := ui.PollEvents()
+	for {
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+		switch e.ID {
+		// if user input q or control-c, exit the program.
+		case "q", "<C-c>":
+			ui.Close()
```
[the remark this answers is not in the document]
rminnich: why not defer ui.Close()

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+				warning.Text = ""
+				ui.Render(warning)
+			}
+			input.Text += e.ID
+			ui.Render(input)
+		}
+	}
+}
+
+// NewInputWindow will create a new ui window and display an input box. We have this function because GetInput function will display in the current window, and sometimes we want anew window to make the UI looks clean and neat.
```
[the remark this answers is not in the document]
rminnich: and if not 80, at least SOMETHING less than 400 :-)

cmds/webboot/webboot.go
pallaud: Maybe the way to deal with this is to modify the Entry interface to also have a Do() method like you used to have before, so that you can call Do() directly from the menu function. So you'd probably need to do something like

```
func (u ISO) Do() error {
	err := getKernel(u)
        // Add logic for actually booting the entry
}
```

pkg/menus/menus.go
xsy1996: I am using vim now, but it is not convenient for me. I wondered will VS code be a good choice?

pkg/menus/menus.go
xsy1996: but If I put extra information like path to iso in the entry, other kind of option may nor have this attribute like download option. and download option will need other attribute like download link etc. that will make entry struct too redundancy I guess

xsy1996: > I would argue you need to put more information in the entry, if it does not have enough information.

but If I put extra information like path to iso in the entry, other kind of option may nor have this attribute like download option. and download option will need other attribute like download link etc. that will make entry struct too redundancy I guess

xsy1996: @rminnich

cmds/webboot/webboot.go
xsy1996: I thought about this but the problem is if I run getKernel() in Do(). I don't know how to inform my main func what is getKernel() return. I may need the path to kernel return by getKernel() in next step to replace the kernel in main. And I also think about let Do() return what getKernel() return, but the problem is, in the Entry struct I have declare the format of Do(), but different kind of option (like iso kind and download kind) will want different kind of information so I can not define the format of Do() clearly.

pallaud: So I think the idea here is that Do() should take care of all the steps that happen after an ISO has been selected. This includes finding the kernel and initrd (in our case, this will be done using FindDevices from u-root), and kexec. The only reason why you'd need to go back to your main function is if there was an error, which can be dealt with by returning the error. So this would probably look like

```
type Entry interface {
   Label() String
   IsDefault() bool
  Do() error
}

func (u ISO) Do() error {
	err := getKernel(u)
        // Add logic for actually booting the entry
}

func (u ISO) Do() error {
	err := getKernel(u)
        // Add logic for actually booting the entry
}
```

Does that work?

pkg/menus/menus.go
pallaud: See my other comment for one potential way to resolve this

cmds/webboot/webboot.go
xsy1996: Thank you! I will try this way!

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,307 @@
...
+	if *v {
+		verbose = log.Printf
+	}
+
+	var isos []ISO
+
+	// If isoDir is not given, that means we need use findDevice to find the USB disk
+	// else we walk through the given folder to find all iso inside.
+	if *isoDir == "" {
+
```
pallaud, line 290: Just have this be

```
if len(*isoDir) > 0 {
   // walkfunc
} 
```

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,307 @@
...
+	/*
+	* TODO: code this function after FindDevice is ready
+	*/
+	return nil
+}
+
+// isoMenu is to  build the iso menu page and return the iso path
+// and kernel path of which user choose.
+func isoMenu(isos []ISO) error {
```
pallaud, line 196: Instead of isoMenu can this be webbootMenu or mainMenu or something?

pkg/menus/menus.go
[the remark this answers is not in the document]
pallaud: Reminder to make this const!

pkg/menus/menus.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Entry contains all the information needed for a boot entry
+type Entry interface {
+	// Label is the string will show in menu
+	Label() string
+	// IsDefault is true means this entry will be hit by default.
+	// If there is many default choise, the first on in the list will be choose.
+	// the rest will be ignored.
+	IsDefault() bool
+	Do() error
```
pallaud, line 22: Add a 1 line comment on what Do() is for

pkg/menus/menus.go
pallaud: Reminder to start the menu numbering from 0!

pkg/menus/menus.go
```diff
@@ -0,0 +1,174 @@
...
+	// design the size and location of the menu
+	l.SetRect(0, location, menuWidth, location+len(entries)+2)
+	location += len(entries) + 2
+	l.TextStyle.Fg = ui.ColorWhite
+	ui.Render(l)
+
+	// checkValidFunc is a function for GetInput to check the format of input.
+	// Since we want user to choose a option from the menu, the checkValidFunction will check :
+	// 1.input is a number; 2.input number does not exceed the number of options.
+	checkValidFunc := func(input string) (bool, string) {
```
pallaud, line 52: Reminder: in Go shorter names are better. Can you change the name of this function to be isValid?

pkg/menus/menus.go
pallaud: Ping on this!

pkg/menus/menus.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		c, err := strconv.Atoi(input)
+		if err != nil || c < 1 || c > len(entries) {
+			return false, "your input is not a valid entry number."
+		}
+		return true, ""
+	}
+
+	// call GetInput to get user's choice
+	choose, err := GetInput(introwords, location, 100, 1, checkValidFunc)
```
pallaud, line 69: Can we rename this 'choice'?

pkg/menus/menus.go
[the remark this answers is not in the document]
pallaud: Ping!

pkg/menus/menus.go
```diff
@@ -0,0 +1,162 @@
...
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+		switch e.ID {
+		// if user input q or control-c, exit the program.
+		case "q", "<C-c>":
+			ui.Close()
+			ui.Close()
+			os.Exit(1)
```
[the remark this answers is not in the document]
pallaud: Ping!

pkg/menus/menus.go
```diff
@@ -0,0 +1,174 @@
...
+			input.Text += e.ID
+			ui.Render(input)
+		}
+	}
+}
+
+// NewInputWindow will create a new ui window and display an input box.
+// We have this function because GetInput function will display in the current window.
+// And sometimes we want anew window to make the UI looks clean and neat.
+func NewInputWindow(introwords string, wid int, ht int, checkValidFunc func(string) (bool, string)) (string, error) {
```
pallaud, line 167: It seems that you always call NewInputWindow with the same wid and ht. What do you think of removing those arguments and doing something like

```
func NewInputWindow(desc string, isValid func(string) (bool, string)) (string, error) {
  return NewCustomInputWindow(desc, 100, 1, isValid)
}

func NewCustomInputWindow(desc string, wid, ht int, isValid func(string) (bool, string)) (string, error) {
}
```

so that way you don't need to  keep inputting widt/ht if you can use the default, and you have the option to have a custom size if you want

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,307 @@
...
+func main() {
+	flag.Parse()
+
+	if *v {
+		verbose = log.Printf
+	}
+
+	var isos []ISO
+
+	// If isoDir is not given, that means we need use findDevice to find the USB disk
```
pallaud, line 287: You can just add a comment here saying TODO: support finding ISOs from bootable device

pkg/menus/menus.go
```diff
@@ -0,0 +1,174 @@
...
+	Label() string
+	// IsDefault is true means this entry will be hit by default.
+	// If there is many default choise, the first on in the list will be choose.
+	// the rest will be ignored.
+	IsDefault() bool
+	Do() error
+}
+
+// DisplayMenu will present all entries into a menu with numbers, so that user can choose from them.
+func DisplayMenu(menuTitle string, introwords string, location int, entries []Entry) (Entry, error) {
```
pallaud, line 26: Note: for future, it'd be nice to be able to go 'back' in this menu flow, maybe we can open an issue for that and get to it later

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,307 @@
...
+	url     string
+}
+
+// DownloadByLinkOption contains information of the iso user want to download,
+// include it's name and the download URL
+type DownloadByLinkOption struct {
+	isoName string
+	url     string
+}
+
```
rminnich, line 73: downloadoption and downloadbylinkoption are two nearly identical structs. If you created an interface called BootMedia, it could have several methods, 
type BootMedia interface {
IsDefault() bool
Name() string
Path() string // This can be a URL, a directory, or whatever
Kernel() string
}

This is just an idea. See what you think.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,307 @@
...
+func (u DownloadOption) IsDefault() bool {
+	return true
+}
+
+func (u DownloadOption) String() string {
+	return fmt.Sprintf("%+v\n", u)
+}
+
+// Label is the string this entry displays in the menu page.
+// Download option will be displayed as "Download" in the menu page.
```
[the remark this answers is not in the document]
xsy1996: yes, golint ask me to add some comments here

xsy1996:
the changes will be added in a series of commits in another PR.

PR #94, opened 2020-07-10
Add menu pkg

xsy1996:
Create menu pkg and add getting user input feature

### DIFF
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -1,20 +1,83 @@
 package menu
 
 import (
-    ui "github.com/gizak/termui/v3"
-    "github.com/gizak/termui/v3/widgets"
+	"os"
+
+	ui "github.com/gizak/termui/v3"
+	"github.com/gizak/termui/v3/widgets"
 )
 
 const menuWidth = 50
 
 // newParagraph will return a paragraph object with given initial text.
-func newParagraph(initText string, Border bool, location int, wid int, ht int) *widgets.Paragraph{
-    p := widgets.NewParagraph()
-    p.Text = initText
-    p.Border = false
-    p.SetRect(0, location, wid, location+ht)
-    p.TextStyle.Fg = ui.ColorWhite
-    ui.Render(p)
-    return p
+func newParagraph(initText string, border bool, location int, wid int, ht int) *widgets.Paragraph {
+	p := widgets.NewParagraph()
+	p.Text = initText
+	p.Border = border
+	p.SetRect(0, location, wid, location+ht)
+	p.TextStyle.Fg = ui.ColorWhite
+	ui.Render(p)
+	return p
 }
 
+// GetInput will present an input box to user and return the user's input.
+// GetInout will check validation of input using isValid function.
+func GetInput(introwords string, location int, wid int, ht int, isValid func(string) (bool, string)) (string, error) {
+	// intro paragraph is to tell user what need to be input here
+	newParagraph(introwords, false, location, len(introwords)+4, 3)
+	location += 2
+
+	// input paragraph is where user input
+	input := newParagraph("", true, location, wid, ht+2)
+	location += ht + 2
+
+	// warning paragraph is to warn user their input is not valid
+	warning := newParagraph("", false, location, wid, 3)
+	ui.Render(warning)
+
+	// keep tracking all input from user
+	uiEvents := ui.PollEvents()
+	for {
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+        switch e.ID {
+		// if user input q or control-c, exit the program.
+		case "q", "<C-c>":
+			ui.Close()
+			os.Exit(1)
+        // If user hit enter means he did his choice.
+		// So check whether the input is valid or not by isValid function.
+		// If isValid is nil, just return the text anyway.
+		// If the input is not acceptable, show a warning.
+		case "<Enter>":
+            if isValid == nil {
+				return input.Text, nil
+			}
+			valid, warningWords := isValid(input.Text)
+			// if input is vilid, directly return the input itself.
+			// else we should clear the input box and show warnings.
+			if valid {
+				return input.Text, nil
+			}
+			input.Text = ""
+			warning.Text = warningWords
+			ui.Render(input)
+			ui.Render(warning)
+		//If userhit backspace, remove the last input character
+		case "<Backspace>":
+			input.Text = input.Text[:len(input.Text)-1]
+			ui.Render(input)
+		// as long as user do not hit enter or q or control-c, assuming that user is still inputing his choice
+		default:
+			// clear warning when user input
+			if warning.Text != "" {
+				warning.Text = ""
+				ui.Render(warning)
+			}
+			input.Text += e.ID
+			ui.Render(input)
+		}
+	}
+}
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -1,0 +1,20 @@
+package menu
+
+import (
+    ui "github.com/gizak/termui/v3"
+    "github.com/gizak/termui/v3/widgets"
+)
+
+const menuWidth = 50
+
+// newParagraph will return a paragraph object with given initial text.
+func newParagraph(initText string, Border bool, location int, wid int, ht int) *widgets.Paragraph{
+    p := widgets.NewParagraph()
+    p.Text = initText
+    p.Border = false
+    p.SetRect(0, location, wid, location+ht)
+    p.TextStyle.Fg = ui.ColorWhite
+    ui.Render(p)
+    return p
+}
+
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -1,7 +1,7 @@
 package menu
 
 import (
-	"os"
+	"io"
 
 	ui "github.com/gizak/termui/v3"
 	"github.com/gizak/termui/v3/widgets"
@@ -16,62 +16,35 @@ func newParagraph(initText string, border bool, location int, wid int, ht int) *
 	p.Border = border
 	p.SetRect(0, location, wid, location+ht)
 	p.TextStyle.Fg = ui.ColorWhite
-	ui.Render(p)
 	return p
 }
 
-// GetInput will present an input box to user and return the user's input.
-// GetInout will check validation of input using isValid function.
-func GetInput(introwords string, location int, wid int, ht int, isValid func(string) (bool, string)) (string, error) {
-	// intro paragraph is to tell user what need to be input here
-	newParagraph(introwords, false, location, len(introwords)+4, 3)
-	location += 2
-
-	// input paragraph is where user input
-	input := newParagraph("", true, location, wid, ht+2)
-	location += ht + 2
-
-	// warning paragraph is to warn user their input is not valid
-	warning := newParagraph("", false, location, wid, 3)
-	ui.Render(warning)
-
+func processInput(input, warning *widgets.Paragraph, isValid func(string) (string, bool), uiEvents <-chan ui.Event) (string, error) {
 	// keep tracking all input from user
-	uiEvents := ui.PollEvents()
 	for {
 		e := <-uiEvents
 		if e.Type != ui.KeyboardEvent {
 			continue
 		}
-        switch e.ID {
-		// if user input q or control-c, exit the program.
-		case "q", "<C-c>":
-			ui.Close()
-			os.Exit(1)
-        // If user hit enter means he did his choice.
-		// So check whether the input is valid or not by isValid function.
-		// If isValid is nil, just return the text anyway.
-		// If the input is not acceptable, show a warning.
+		switch e.ID {
+		case "<C-d>":
+			return "", io.EOF
 		case "<Enter>":
-            if isValid == nil {
-				return input.Text, nil
-			}
-			valid, warningWords := isValid(input.Text)
-			// if input is vilid, directly return the input itself.
-			// else we should clear the input box and show warnings.
-			if valid {
+			warningWords, ok := isValid(input.Text)
+			if ok {
 				return input.Text, nil
 			}
 			input.Text = ""
 			warning.Text = warningWords
 			ui.Render(input)
 			ui.Render(warning)
-		//If userhit backspace, remove the last input character
 		case "<Backspace>":
+			if len(input.Text) == 0 {
+				continue
+			}
 			input.Text = input.Text[:len(input.Text)-1]
 			ui.Render(input)
-		// as long as user do not hit enter or q or control-c, assuming that user is still inputing his choice
 		default:
-			// clear warning when user input
 			if warning.Text != "" {
 				warning.Text = ""
 				ui.Render(warning)
@@ -80,4 +53,21 @@ func GetInput(introwords string, location int, wid int, ht int, isValid func(str
 			ui.Render(input)
 		}
 	}
+}
+
+// GetInput will present an input box to user and return the user's input.
+// GetInput will check validation of input using isValid function.
+func GetInput(introwords string, location int, wid int, ht int, isValid func(string) (string, bool)) (string, error) {
+	intro := newParagraph(introwords, false, location, len(introwords)+4, 3)
+	location += 2
+	input := newParagraph("", true, location, wid, ht+2)
+	location += ht + 2
+	warning := newParagraph("", false, location, wid, 3)
+
+	ui.Render(intro)
+	ui.Render(input)
+	ui.Render(warning)
+
+	uiEvents := ui.PollEvents()
+	return processInput(input, warning, isValid, uiEvents)
 }
--- a/pkg/menu/menu_test.go
+++ b/pkg/menu/menu_test.go
@@ -1,15 +1,35 @@
 package menu
 
 import (
-    "testing"
-    ui "github.com/gizak/termui/v3"
+	ui "github.com/gizak/termui/v3"
+	"strconv"
+	"testing"
 )
 
-func Test_newParagraph(t *testing.T){
-    if err := ui.Init(); err != nil {
-        t.Log(err)
-    }
-    defer ui.Close()
-    p := newParagraph("test", false, 0, 50, 3)
-    t.Log(p.Text)
+func Test_newParagraph(t *testing.T) {
+	if err := ui.Init(); err != nil {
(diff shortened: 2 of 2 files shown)


pkg/menu/menu.go
```diff
@@ -0,0 +1,83 @@
...
+    p.Border = border
+    p.SetRect(0, location, wid, location+ht)
+    p.TextStyle.Fg = ui.ColorWhite
+    ui.Render(p)
+    return p
+}
+
+// GetInput will present an input box to user and return the user's input.
+// GetInout will check validation of input using isValid function.
+func GetInput(introwords string, location int, wid int, ht int, isValid func(string) (bool, string)) (string, error) {
```
rminnich, line 25: just checking, but did you run gofmt -w on these files.

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func Test_newParagraph(t *testing.T){
+    if err := ui.Init(); err != nil {
+        t.Log(err)
+    }
+    defer ui.Close()
+    p := newParagraph("newParagraph test", false, 0, 50, 3)
+    t.Log(p.Text)
+}
```
rminnich, line 16: How do you know it worked? The test needs to be in this code. It seems you are look at p.Text for some indication. Can you compare p.Text to a string and see if it has an expected value?

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,36 @@
+package menu
+
+import (
+    "strconv"
+    "testing"
+    ui "github.com/gizak/termui/v3"
+)
+
+func Test_newParagraph(t *testing.T){
+    if err := ui.Init(); err != nil {
```
rminnich, line 10: If this fails, why would use use t.Log instead of t.Fatal?

pkg/menu/menu_test.go
rminnich: remember, when you run a test, it must be able to run as part of an unattended test, where you do not watch it run. It will be run by constant integration framework, for example. So your test has to have its own checks to make sure it worked. Just printing things via t.Log is not enough.

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,36 @@
...
+        t.Log(err)
+    }
+    defer ui.Close()
+    p := newParagraph("newParagraph test", false, 0, 50, 3)
+    t.Log(p.Text)
+}
+
+func Test_GetInput(t *testing.T){
+    if err := ui.Init(); err != nil {
+        t.Log(err)
```
rminnich, line 20: I think you should be calling t.Fatal

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,36 @@
...
+		if err != nil || c<0 {
+			return false, "your input is not a valid entry number."
+		}
+		return true, ""
+	}
+    input, err := GetInput("GetInput test", 0, 50, 1, isValid)
+    t.Log(input)
+    if(err!=nil){
+        t.Errorf("%v",err)
+    }
```
rminnich, line 34: how did GetInput get its input? If you had to type into it, then this is not a complete test. It must be able to run without user input.

pkg/menu/menu.go
```diff
@@ -0,0 +1,83 @@
...
+	// intro paragraph is to tell user what need to be input here
+	newParagraph(introwords, false, location, len(introwords)+4, 3)
+	location += 2
+
+	// input paragraph is where user input
```
pallaud, line 30: I think for cases like this, the variable name is enough to explain what it is for, and the comment isn't needed. If your code is self-explanatory, then you need fewer comments, your code is cleaner :)

pallaud: Try to find other places where this might be the case

pkg/menu/menu.go
```diff
@@ -0,0 +1,83 @@
...
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+        switch e.ID {
+		// if user input q or control-c, exit the program.
+		case "q", "<C-c>":
+			ui.Close()
+			os.Exit(1)
+        // If user hit enter means he did his choice.
```
pallaud, line 50: Nit: comment alignment

pkg/menu/menu.go
```diff
@@ -0,0 +1,83 @@
...
+		// So check whether the input is valid or not by isValid function.
+		// If isValid is nil, just return the text anyway.
+		// If the input is not acceptable, show a warning.
+		case "<Enter>":
+            if isValid == nil {
```
pallaud, line 55: Nit: alignment is off. Gofmt should fix that for you?

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	defer ui.Close()
+	isValid := func(input string) (bool, string) {
+        c, err := strconv.Atoi(input)
+		if err != nil || c < 0 {
+			return false, "your input is not a valid entry number."
+		}
+		return true, ""
+	}
+	input, err := GetInput("GetInput test", 0, 50, 1, isValid)
```
pallaud, line 30: What Ron said earlier applies here as well, how would you check that this worked without having to watch it? So basically, how can you simulate the keyboard events for input here?

pkg/menu/menu.go
```diff
@@ -0,0 +1,75 @@
...
+	p.Border = border
+	p.SetRect(0, location, wid, location+ht)
+	p.TextStyle.Fg = ui.ColorWhite
+	ui.Render(p)
+	return p
+}
+
+// GetInput will present an input box to user and return the user's input.
+// GetInout will check validation of input using isValid function.
+func GetInput(introwords string, location int, wid int, ht int, isValid func(string) (bool, string), testMode bool, testInput chan ui.Event) (string, error) {
```
pallaud, line 25: Suggestion: what if you had your functions set up like this

```
func GetInput(introwords string, location int, wid int, ht int, isValid func(string) (bool, string)) (string, error) {
   // show intro text, set up input and warning paragraphs
   return processInput(input, warning, isValid, ui.PollEvents())
}

func processInput(input, warning *widgets.Paragraph, isValid func(string) (bool, string), ch chan ui.Event) (string, error) {
   // have the for loop that reads from ch 
}
```
Then in your test, you can set up the channel using ui.PollEvents() and pass that channel to processInput, and then send ui events on it. Would that work?

pkg/menu/menu.go
```diff
@@ -0,0 +1,75 @@
...
+	p.Text = initText
+	p.Border = border
+	p.SetRect(0, location, wid, location+ht)
+	p.TextStyle.Fg = ui.ColorWhite
+	ui.Render(p)
+	return p
+}
+
+// GetInput will present an input box to user and return the user's input.
+// GetInout will check validation of input using isValid function.
```
pallaud, line 24: Nit: spelling of GetInput

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func Test_newParagraph(t *testing.T) {
+	if err := ui.Init(); err != nil {
+		t.Fatal(err)
+	}
+	defer ui.Close()
+	const testText = "newParagraph test"
+	p := newParagraph(testText, false, 0, 50, 3)
+	if testText != p.Text {
+		t.Errorf("p.Text is %s, expect %s", p.Text, testText)
```
pallaud, line 17: Nit: I think convention for errors in Go tests is something like `t.Errorf("incorrect value for p.Text. got: %v, want: %v", p.Text, testText)`

pkg/menu/menu.go
pallaud: I think this might also allow you to test that the warnings are being correctly rendered, because if you pass the `warning` *widgets.Paragraph and pass an incorrect input, you can check that the warning text matches what you expect right?

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,66 @@
...
+		"1", "0", "0", "a", "<Enter>",
+		"1", "0", "a", "<Backspace>", "0", "<Enter>"})
+	isValid := func(input string) (bool, string) {
+		c, err := strconv.Atoi(input)
+		if err != nil || c < 0 {
+			return false, "Input is not a valid entry number."
+		}
+		return true, ""
+	}
+	inputText, err := processInput(input, warning, isValid, ch)
```
pallaud, line 56: Does it not work to use ui.PollEvents() to get the channel here instead of having to make your own function? That was the purpose of separating out the processInput function right?

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"strconv"
+	"testing"
+)
+
+func Test_newParagraph(t *testing.T) {
+	if err := ui.Init(); err != nil {
+		t.Fatal(err)
+	}
+	defer ui.Close()
+	const testText = "newParagraph test"
```
pallaud, line 14: I think this can just be `testTest := "newParagraph test"`

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, id := range input {
+		key = ui.Event{
+			Type: ui.KeyboardEvent,
+			ID:   id,
+		}
+		ch <- key
+	}
+}
+
+func Test_GetInput(t *testing.T) {
```
pallaud, line 32: I think you can have 2 tests, Test_GetInput_Good and Test_GetInput_Bad or something, where the first once just does a basic test that you were doing before, and the second does this new version with bad inputs. That helps separate the testing of the basic functionality from the warning part.

pkg/menu/menu_test.go
xsy1996: yes, I tried. it works too

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,66 @@
...
+	p := newParagraph(testText, false, 0, 50, 3)
+	if testText != p.Text {
+		t.Errorf("incorrect value for p.Text. got: %v, want: %v", p.Text, testText)
+	}
+}
+
+func pressKey(ch chan ui.Event, input []string) {
+	var key ui.Event
+	for _, id := range input {
+		key = ui.Event{
```
pallaud, line 24: I think you can just do 
```
key = ui.Event{
  // fill this in
}
```

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		ch <- key
+	}
+}
+
+func Test_GetInput(t *testing.T) {
+	if err := ui.Init(); err != nil {
+		t.Fatal(err)
+	}
+	defer ui.Close()
```
pallaud, line 36: nit: maybe add some newlines between your lines here and some of the other places where you have a lot of lines - just makes it easier to read :)

pkg/menu/menu.go
```diff
@@ -0,0 +1,73 @@
...
+func processInput(input, warning *widgets.Paragraph, isValid func(string) (bool, string), uiEvents <-chan ui.Event) (string, error) {
+	// keep tracking all input from user
+	for {
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+		switch e.ID {
+		// if user input control-c, exit the program.
+		case "<C-c>":
```
rminnich, line 32: I think C-d would be better, that is the usual character used in Unix meaning no more input.

Also, the convention for programs like this that take input is to return io.Eof, not just exit. The calling program might wish to do cleanup of its own.

pkg/menu/menu.go
```diff
@@ -0,0 +1,72 @@
...
+			if valid {
+				return input.Text, nil
+			}
+			input.Text = ""
+			warning.Text = warningWords
+			ui.Render(input)
+			ui.Render(warning)
+		// If user hit backspace, remove the last input character
+		case "<Backspace>":
+			input.Text = input.Text[:len(input.Text)-1]
```
rminnich, line 48: If the first thing they hit is a backspace, what happens? Does input.Text[:-1] work?

see https://play.golang.org/p/Tc-3E5sEf56

pkg/menu/menu.go
```diff
@@ -0,0 +1,72 @@
...
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+		switch e.ID {
+		// if user input control-d, raise error io.EOF meaning no more input.
+		case "<C-d>":
+			return "", io.EOF
+		case "<Enter>":
+			if isValid == nil {
```
rminnich, line 35: Passing nil pointers around can cause real trouble. 

A common pattern here would be to provide a func:
func alwaysValid(s string) (bool, string) {
return true, s
}

then for the cases where someone might have passed in nil, they can instead pass in alwaysValid
Then you can get rid of lines 35-37 because the function can always be called. 

Also I notice the isValid fund returns a bool and a string, and the processInput returns a string and an error. It is common for validators to return (string, bool) because that is similar to how maps work, e.g. if I have a map[string]string I say
x, ok := map["s"]
if ok {...}
and your isValid could be the same pattern:
isValid(string) (string, bool)
So lines 35-41 become this:
if s, ok := isValid(input.Text); ok {
return s, nil
}

pkg/menu/menu.go
```diff
@@ -0,0 +1,72 @@
...
+			ui.Render(input)
+			ui.Render(warning)
+		// If user hit backspace, remove the last input character
+		case "<Backspace>":
+			input.Text = input.Text[:len(input.Text)-1]
+			ui.Render(input)
+		default:
+			if warning.Text != "" {
+				warning.Text = ""
+				ui.Render(warning)
```
rminnich, line 53: I am not sure I understand this. Do you want to render the warning before you set it to ""?

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,106 @@
...
+	// second input is bad input "100a"
+	// third input is good input "100" but contain a process of typo then backspace
+	// now the warning text should be ""
+	go pressKey(uiEvents, []string{"b", "a", "d", "<Enter>",
+		"1", "0", "0", "a", "<Enter>",
+		"1", "0", "a", "<Backspace>", "0", "<Enter>"})
+
+	isValid := func(input string) (bool, string) {
+		c, err := strconv.Atoi(input)
+		if err != nil || c < 0 {
```
rminnich, line 89: if you use strconv.ParseUInt then you need not check for c < 0

pkg/menu/menu.go
```diff
@@ -0,0 +1,72 @@
...
+const menuWidth = 50
+
+// newParagraph will return a paragraph object with given initial text.
+func newParagraph(initText string, border bool, location int, wid int, ht int) *widgets.Paragraph {
+	p := widgets.NewParagraph()
+	p.Text = initText
+	p.Border = border
+	p.SetRect(0, location, wid, location+ht)
+	p.TextStyle.Fg = ui.ColorWhite
+	ui.Render(p)
```
rminnich, line 19: In general, the UI packages I have seen separate the creation of a thing from rendering that thing. So a newParagraph creates the new paragraph with all the options, but another function would normally render it. This is just a thought and it is up to you.

pkg/menu/menu.go
rminnich: Is there a race condition between using the warning.Text and setting it? You may want to run the go race detector.

xsy1996: I think render is after I set warning to "" since in my mind render is to implement the change. and what you mean about  a race condition ? what I want to do here is to reset the warning when user begin to write a new input. If I set the warning.Text and then render it, will there be a race condition?

rminnich: ah ha. I did not understand that this was clearing the message :-)
I think there's likely no problem with races but just wondered.

pkg/menu/menu.go
```diff
@@ -0,0 +1,75 @@
...
+
+func processInput(input, warning *widgets.Paragraph, isValid func(string) (string, bool), uiEvents <-chan ui.Event) (string, error) {
+	// keep tracking all input from user
+	for {
+		e := <-uiEvents
+		if e.Type != ui.KeyboardEvent {
+			continue
+		}
+		switch e.ID {
+		// if user input control-d, raise error io.EOF meaning no more input.
```
rminnich, line 30: Your code is nice and clear so I doubt you need this comment.

pkg/menu/menu.go
```diff
@@ -0,0 +1,75 @@
...
+		case "<C-d>":
+			return "", io.EOF
+		case "<Enter>":
+			if warningWords, ok := isValid(input.Text); ok {
+				return input.Text, nil
+			} else {
```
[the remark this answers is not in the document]
xsy1996: This is because the warningWords is defined in this if sentence. I will change it to 
warningWords, ok := isValid(input.Text)
if  ok {
	return input.Text, nil
} 
...

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,105 @@
...
+	go pressKey(uiEvents, []string{"t", "e", "s", "t", "<Enter>"})
+
+	isValid := func(input string) (string, bool) {
+		return "", true
+	}
+
+	inputText, err := processInput(input, warning, isValid, uiEvents)
+
+	if err != nil {
+		t.Errorf("Error: %v", err)
```
pallaud, line 55: Maybe here just say `t.Errorf("processInput failed: %v", err)`

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,105 @@
+package menu
+
+import (
+	ui "github.com/gizak/termui/v3"
+	"strconv"
+	"testing"
+)
+
+func Test_newParagraph(t *testing.T) {
```
pallaud, line 9: Nit: I think convention in Go is to not use underscores for naming so you can name your tests TestXxx, so this would become `TestNewParagraph`

pkg/menu/menu_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if inputText != testText {
+		t.Errorf("Incorrect value for input. got: %v, want: %v", inputText, testText)
+	}
+	if warning.Text != "" {
+		t.Errorf("Incorrect value for warning. got: %v, want nothing", warning.Text)
+	}
+}
+
+func Test_GetInput_Complicated(t *testing.T) {
```
pallaud, line 65: Nit: maybe rename to `TestGetInputInvalid` to indicate exactly what functionality youre testing?

PR #96, opened 2020-07-20
add cmds/webboot

xsy1996:
creating a menu with download_by_bookmark and download_by_link options

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -8,7 +8,87 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
+
+	ui "github.com/gizak/termui/v3"
+	"github.com/u-root/webboot/pkg/menu"
 )
+
+const (
+	downloadByLinkLabel = "Download by link"
+	tcURL               = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
+)
+
+// bookmark record the list of iso webboot allow user to download
+var bookmark = map[string]BookMarkISO{
+	"tinycore": BookMarkISO{
+		isDefault: false,
+		url:       tcURL,
+		label:     "Download Tinycore v10.1",
+		isoName:   "TinyCorePure64-10.1.iso",
+	},
+}
+
+// BookMarkISO contains information of the iso user want to download
+// expected to implement Entry interface
+type BookMarkISO struct {
+	isDefault bool
+	url       string
+	label     string
+	isoName   string
+}
+
+// CachedISO contains information of the iso cached in the memory
+// expected to implement Entry interface
+type CachedISO struct {
+	isDefault bool
+	label     string
+	path      string
+	isoName   string
+	// todo: information parsed from config file
+}
+
+// DownloadByLink is to implement "Download by link" option in the menu
+type DownloadByLink struct{}
+
+// Label is the string this iso displays in the menu page.
+func (d *DownloadByLink) Label() string {
+	return "Download by link"
+}
+
+// IsDefault is for mark whether this iso is a default choice.
+// assume that DownloadByLink will not be a default option.
+func (d *DownloadByLink) IsDefault() bool {
+	return false
+}
+
+// DownloadByBookmark is to implement "Download by bookmark" option in the menu
+type DownloadByBookmark struct{}
+
+// Exec performs the following process after the entry is chosen
+func (d *DownloadByLink) Exec(uiEvents <-chan ui.Event) error {
+	link, err := menu.NewInputWindow("Please input the link", menu.AlwaysValid, uiEvents)
+	if err != nil {
+		return err
+	}
+	validIsoName := func(input string) (string, string, bool) {
+		re := regexp.MustCompile(`[\w]+.iso`)
+		if re.Match([]byte(input)) {
+			return input, "", true
+		}
+		return "", "File name should only contain [a-zA-Z0-9_], and should end in .iso", false
+	}
+	filename, err := menu.NewInputWindow("Input ISO name", validIsoName, uiEvents)
+	if err != nil {
+		return err
+	}
+	fPath := filepath.Join("/tmp", filename)
+	if err := download(link, fPath); err != nil {
+		return err
+	}
+	// todo: boot the iso
+	return nil
+}
 
 func linkOpen(URL string) (io.ReadCloser, error) {
 	u, err := url.Parse(URL)
@@ -32,15 +112,14 @@ func linkOpen(URL string) (io.ReadCloser, error) {
 	return nil, fmt.Errorf("%q: linkopen only supports file://, https://, and http:// schemes", URL)
 }
 
-// downloadIso will download a file from URL and save it as desDir/filename
-func download(URL, desDir, filename string) error {
+// download will download a file from URL and save it as desDir/filename
+func download(URL, fPath string) error {
 
 	isoReader, err := linkOpen(URL)
 	if err != nil {
 		return err
 	}
 	defer isoReader.Close()
-	fPath := filepath.Join(desDir, filename)
 	f, err := os.Create(fPath)
 	if err != nil {
 		return err
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -17,15 +17,22 @@ import (
 const (
 	downloadByLinkLabel = "Download by link"
 	tcURL               = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
+	wbtcpURL            = "https://github.com/u-root/webboot-distro/raw/master/iso/tinycore/10.x/x86_64/release/TinyCorePure64.iso"
 )
 
 // bookmark record the list of iso webboot allow user to download
-var bookmark = map[string]BookMarkISO{
-	"tinycore": BookMarkISO{
+var bookmark = []*BookMarkISO{
+	&BookMarkISO{
 		isDefault: false,
 		url:       tcURL,
 		label:     "Download Tinycore v10.1",
 		isoName:   "TinyCorePure64-10.1.iso",
+	},
+	&BookMarkISO{
+		isDefault: false,
+		url:       wbtcpURL,
+		label:     "Webboot Tinycorepure",
+		isoName:   "TinyCorePure64.iso",
 	},
 }
 
@@ -38,6 +45,28 @@ type BookMarkISO struct {
 	isoName   string
 }
 
+// Label is the string this iso displays in the menu page.
+func (b *BookMarkISO) Label() string {
+	return b.label
+}
+
+// IsDefault is for mark whether this iso is a default choice.
+func (b *BookMarkISO) IsDefault() bool {
+	return b.isDefault
+}
+
+// Exec performs the following process after the entry is chosen
+func (b *BookMarkISO) Exec(_ <-chan ui.Event) error {
+	link := b.url
+	filename := b.isoName
+	fPath := filepath.Join("/tmp", filename)
+	if err := download(link, fPath); err != nil {
+		return err
+	}
+	// todo: boot the iso
+	return nil
+}
+
 // CachedISO contains information of the iso cached in the memory
 // expected to implement Entry interface
 type CachedISO struct {
@@ -46,6 +75,36 @@ type CachedISO struct {
 	path      string
 	isoName   string
 	// todo: information parsed from config file
+}
+
+// DownloadByBookmark is to implement "Download by bookmark" option in the menu
+type DownloadByBookmark struct{}
+
+// Label is the string this iso displays in the menu page.
+func (d *DownloadByBookmark) Label() string {
+	return "Download by link"
+}
+
+// IsDefault is for mark whether this iso is a default choice.
+// assume that DownloadByLink will not be a default option.
+func (d *DownloadByBookmark) IsDefault() bool {
+	return false
+}
+
+// Exec performs the following process after the entry is chosen
+func (d *DownloadByBookmark) Exec(uiEvents <-chan ui.Event) error {
+	entries := []menu.Entry{}
+	for _, e := range bookmark {
+		entries = append(entries, e)
+	}
+
+	chosen, err := menu.DisplayMenu("Bookmarks", "Input your choice", entries, uiEvents)
+	if err != nil {
+		return err
+	}
+
+	return chosen.Exec(uiEvents)
+	// todo: boot the iso
 }
 
 // DownloadByLink is to implement "Download by link" option in the menu
@@ -61,9 +120,6 @@ func (d *DownloadByLink) Label() string {
 func (d *DownloadByLink) IsDefault() bool {
 	return false
 }
-
-// DownloadByBookmark is to implement "Download by bookmark" option in the menu
-type DownloadByBookmark struct{}
 
 // Exec performs the following process after the entry is chosen
 func (d *DownloadByLink) Exec(uiEvents <-chan ui.Event) error {
@@ -98,7 +154,7 @@ func linkOpen(URL string) (io.ReadCloser, error) {
 	switch u.Scheme {
 	case "file":
 		return os.Open(URL[7:])
-	case "http", "https:":
+	case "http", "https":
 		resp, err := http.Get(URL)
 		if err != nil {
 			return nil, err
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -10,120 +10,39 @@ import (
 	"path/filepath"
 	"regexp"
 
-	ui "github.com/gizak/termui/v3"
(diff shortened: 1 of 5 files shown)


cmds/webboot/webboot.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func linkOpen(URL string) (io.ReadCloser, error) {
```
rminnich, line 12: I think it might be good to look at the URL package, it's pretty good.

https://play.golang.org/p/P-WuEwxoP41

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
+package main
```
pallaud, line 1: I'm thinking looking at the current code that if we keep everything in one file, it might get too long and unwieldy - what if we separated it into a header.go (rename to something better) which has type declarations and bookmarks, etc., and a webboot.go, which has the actual main function and logic we want? Or maybe even a separate package, not sure, @rminnich any thoughts?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// Label is the string this iso displays in the menu page.
+func (d *DownloadedDistro) Label() string {
+	return d.label
+}
+
+// Exec performs the following process after the entry is chosen
+func (d *DownloadedDistro) Exec(uiEvents <-chan ui.Event) error {
+	if d.Label() == downloadByLinkLabel {
```
pallaud, line 77: If we already know that this is a DownloadedDistro why do we need to check the label?

pallaud: Also note for future, if you find yourself having large blocks of code in an if statement, you can instead do something like

```
if (d.Label() != downloadByLinkLabel) {
    return nil
}
// rest of code
```
to keep it easier to read

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+		noConstraint := func(input string) (string, string, bool) {
```
pallaud, line 78: I think its nicer to have the names be a little clearer about what the function is doing. Maybe spaceConstraint could instead be checkValidIso? And I think Ron mentioned this at some point, but maybe have a function alwaysValid or something defined in your menu pkg, and then you can do `menu.NewInputWindow("Input URL:", menu.AlwaysValid)` instead of always having to make a function that returns true

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+		link, err := menu.NewInputWindow("Please input the link", noConstraint, uiEvents)
+		if err != nil {
+			return err
+		}
+		spaceConstraint := func(input string) (string, string, bool) {
+			re := regexp.MustCompile(`[\w]+.iso`)
+			if re.Match([]byte(input)) {
+				return input, "", true
+			}
+			return "", "File name should only contain [a-zA-Z0-9_], and should be end as .iso", false
```
pallaud, line 90: Nit: "should end in .iso"

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		filename, err := menu.NewInputWindow("Please input the name of iso", spaceConstraint, uiEvents)
```
pallaud, line 92: As much as its nice to have polite messages, I think it might be better to have short ones :) Something like "Input ISO name:" would work as well!

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("HTTP Get failed: %v", resp.StatusCode)
+		}
+		return resp.Body, nil
+	}
+	return nil, fmt.Errorf("%q: linkopen only supports file://, https://, and http:// schemes", URL)
+}
+
+// downloadIso will download a file from URL and save it as desDir/filename
+func download(URL, fPath string) error {
```
pallaud, line 139: Should this be named downloadIso (that's what the comment says)?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+	Name() string
+	// todo: may have Kernel(), Config() etc later
+}
+
+// DownloadedDistro contains information of the iso user want to download
+type DownloadedDistro struct {
+	isDefault bool
+	label     string
+	isoName   string
+	group     string
```
pallaud, line 46: What is 'group' for?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+// CachedDistro contains information of the iso cached in the memory
+type CachedDistro struct {
+	isDefault bool
+	label     string
+	group     string
+	path      string
+	isoName   string
+}
+
+// GoBack is the backward option
```
pallaud, line 60: I think that instead of building a back option into the webboot cmd, you'd want to build it into your menu pkg. But maybe lets remove this for now, and wait until we've got the basic functionality working before thinking about the back option.

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+
+// BootMedia contains required information of a bootable iso
+// no matter it is cached or downloaded
+type BootMedia interface {
+	Path() string
+	Name() string
+	// todo: may have Kernel(), Config() etc later
+}
+
+// DownloadedDistro contains information of the iso user want to download
```
pallaud, line 41: Maybe add to the comment which interfaces you expect this to apply?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,157 @@
...
+type DownloadedDistro struct {
```
pallaud, line 42: So looking at your structs, I still see a lot of similar fields. The goal with an interface is to reduce the duplication, right? Is there a way you can reduce the duplication here - for example, do you need the isDefault and label fields if you can already have methods for them from an interface?

Another question to consider is why DownloadedDistro needs its own struct. It seems like after the Exec() function executes, you won't use the url again, in which case there's no difference between that and CachedDistro, right? If I remember, you added this because you had 2 options, Download by bookmark and Download by link. Would it make sense to do something like this

```
type DownloadByBookmark struct{}

func (d *DownloadByBookmark) Label() {
    return "Download by bookmark"
}

func (d *DownloadByBookmark) Exec() {
    // Present a menu of the bookmarks, boot the one the user selects
}

type DownloadByLink struct{}

func (d *DownloadByLink) Label() {
    return "Download by link"
}

func (d *DownloadByLink) Exec() {
    // Ask for a link, download and boot
}
```
since you don't really need any information in the struct, since the user will be providing it to you?

cmds/webboot/webboot.go
xsy1996: the reason to check the label is to check if it is a download by link option or a download by bookmark option.since download by link will as k user to input the link and name while download by bookmark do not

cmds/webboot/webboot.go
xsy1996: Maybe I can move the defintion of CachedDistro and DownloadedDistro  to menu pkg but indifferent .go file? Then the webboot.go just call menu pkg for everything?

cmds/webboot/webboot.go
xsy1996: it's for hierarchy menu

cmds/webboot/webboot.go
pallaud: Would my suggestion about having type DownloadByBookmark and type DownloadByLink help solve this?

cmds/webboot/webboot.go
xsy1996: Get you, But then I still need another struct to implement the bookmark option.

xsy1996: and I can not get a BootMedia interface for the download iso in the main program in this way right? because I have no attribute to store these information. All following process should be done in the Exec()

xsy1996: Or we let the all following process happens Exec() anyway, and the BootMedia just a interface that provide Gabriel information for booting. So use this interface Gabriel could ignore the difference between cachediso and downloadediso

xsy1996: I send a schematic diagram in the Chat

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// IsDefault is for mark whether this iso is a default choice.
+func (b *BookMarkISO) IsDefault() bool {
+	return b.isDefault
+}
+
+// Exec performs the following process after the entry is chosen
+func (b *BookMarkISO) Exec(_ <-chan ui.Event) error {
+	link := b.url
```
pallaud, line 60: Since these variables are only used once, maybe you could directly pass b.url and b.isoName to the functions?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+	filename := b.isoName
+	fPath := filepath.Join("/tmp", filename)
+	if err := download(link, fPath); err != nil {
+		return err
+	}
+	// todo: boot the iso
```
pallaud, line 66: For now could you print a message with info about the iso, here and in the other places?

cmds/webboot/webboot.go
pallaud: Could we do what Ron suggested and move some of this into a separate file, types.go or something?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+// BookMarkISO contains information of the iso user want to download
+// expected to implement Entry interface
+type BookMarkISO struct {
+	isDefault bool
+	url       string
+	label     string
+	isoName   string
+}
+
+// Label is the string this iso displays in the menu page.
```
pallaud, line 48: @rminnich having comments for each of these functions feels a little cluttered since they aren't adding any meaning, will the linter complain if we remove them?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+	return nil
+}
+
+// CachedISO contains information of the iso cached in the memory
+// expected to implement Entry interface
+type CachedISO struct {
+	isDefault bool
+	label     string
+	path      string
+	isoName   string
```
pallaud, line 76: Just 'name' is fine!

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+	},
+}
+
+// BookMarkISO contains information of the iso user want to download
+// expected to implement Entry interface
+type BookMarkISO struct {
+	isDefault bool
+	url       string
+	label     string
+	isoName   string
```
pallaud, line 45: Just 'name' is fine!

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+func (b *BookMarkISO) Label() string {
+	return b.label
+}
+
+// IsDefault is for mark whether this iso is a default choice.
+func (b *BookMarkISO) IsDefault() bool {
+	return b.isDefault
+}
+
+// Exec performs the following process after the entry is chosen
```
pallaud, line 58: Maybe in this comment describe what the function is actually doing, what happens when a bookmark ISO is chosen?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+	isoName   string
+	// todo: information parsed from config file
+}
+
+// DownloadByBookmark is to implement "Download by bookmark" option in the menu
+type DownloadByBookmark struct{}
+
+// Label is the string this iso displays in the menu page.
+func (d *DownloadByBookmark) Label() string {
+	return "Download by link"
```
pallaud, line 85: Should this be "Download by bookmark"?

cmds/webboot/webboot.go
pallaud: Same for the comments for the other Exec functions

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+	isoName   string
+}
+
+// Label is the string this iso displays in the menu page.
+func (b *BookMarkISO) Label() string {
+	return b.label
+}
+
+// IsDefault is for mark whether this iso is a default choice.
+func (b *BookMarkISO) IsDefault() bool {
```
pallaud, line 54: I feel like the isDefault functions are cluttering the file a little - would there be a downside to removing isDefault from the entry and just not having a default?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+	for _, e := range bookmark {
+		entries = append(entries, e)
+	}
+
+	chosen, err := menu.DisplayMenu("Bookmarks", "Input your choice", entries, uiEvents)
+	if err != nil {
+		return err
+	}
+
+	return chosen.Exec(uiEvents)
```
pallaud, line 106: Do you need to do this, won't DisplayMenu call the Exec function for the chosen entry?

cmds/webboot/webboot.go
```diff
@@ -0,0 +1,190 @@
...
+
+// IsDefault is for mark whether this iso is a default choice.
+// assume that DownloadByLink will not be a default option.
+func (d *DownloadByLink) IsDefault() bool {
+	return false
+}
+
+// Exec performs the following process after the entry is chosen
+func (d *DownloadByLink) Exec(uiEvents <-chan ui.Event) error {
+	link, err := menu.NewInputWindow("Please input the link", menu.AlwaysValid, uiEvents)
```
pallaud, line 126: Could we change this message to "Enter URL:" ?

cmds/webboot/webboot_test.go
```diff
@@ -0,0 +1,107 @@
...
+		if err := download("errorlink", "/tmp/test.iso"); err.Error() != expected.Error() {
+			t.Errorf("Error: error msg are wrong, want %+v but get %+v", expected, err)
+		}
+	})
+
+	t.Run("download_tinycore", func(t *testing.T) {
+		fPath := "/tmp/test_tinycore.iso"
+		url := "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
+		if err := download(url, fPath); err != nil {
+			t.Errorf("Error: when downloading, get error %+v", err)
```
pallaud, line 36: I think Go will already print the error in such a way that its clear that it's an error, so to make the error messages shorter, you could do something like `t.Errorf("failed to download: %v, err)` - its clear what the error means, and by having the colon its clear that you're printing the actual error received. You could probably do that for most of the error messages in this file

pkg/menu/menu.go
```diff
@@ -22,7 +22,12 @@ type Entry interface {
 	// If there is many default Entrys, choose the the first in the list.
 	IsDefault() bool
 	// Exec performs the following process after an entry is chosen
-	Exec() error
+	Exec(<-chan ui.Event) error
```
pallaud, line 25: I know that you did this so that it was possible to test, but it feels strange to have this parameter... @rminnich any thoughts on how to test without having to do this? I'll keep thinking about it too

cmds/webboot/webboot.go
xsy1996: no it won't currently. you think it would be better to run Exec in the DisplayMenu?

pallaud: I thought that was already happening, sorry I missed that change - hm I think one of the advantages of having Exec is that DisplayMenu can run it, otherwise why do we need that as part of the Entry interface? I guess its not necessary, but I'd do it from DisplayMenu for convenience. If anyone wants to wait until the function returns, they could always pass an empty Exec function?

cmds/webboot/webboot.go
xsy1996: but I think the the format of the Exec() is fixed? if I give Exec() another parameter like url, then other struct implementing Entry should also have this parameter in their Exec?

cmds/webboot/webboot.go
xsy1996: Yes, golint complained

cmds/webboot/webboot.go
xsy1996: I think it is  only about how user feel. I am not sure will they want a default. I think if the user no nothing about the OS, will it be better to give them a default option? or maybe I can make the first entry in the  menu always be a default? Then we can define the default by changing the order of Entry. what do you think?

pallaud: I think either not having a default or having the first entry be the default is fine.

pkg/menu/menu.go
rminnich: why  not put it in the structs that use it, instead of making it a parameter?

cmds/webboot/headers.go
```diff
@@ -0,0 +1,66 @@
+package main
```
pallaud, line 1: Can you rename this types.go? I think Go doesn't have the header concept.

PR #98, opened 2020-07-25
Make hierarchy menu

xsy1996:
Add cached iso option and make a hierarchy menu

Adding a hierarchy to the menu is to make it easier for users to use. With group options, users can easily locate the ISO they want to boot.

A hierarchy menu means  options will be organized by their group. e.g. the tinycoreV11.01.iso option's group is tinycore, so in the hierarchy menu the user should choose the tinycore group first them meet the tiycoreV11.01.iso option


cmds/webboot/webboot.go
```diff
@@ -108,3 +113,21 @@ func download(URL, fPath string) error {
 	}
 	return nil
 }
+
+func getCachedIsos(cachedDir string) []*CachedISO {
```
pallaud, line 117: At what point is this function called?

xsy1996: I think I will call a function to get all the block device, then mount it one by one, and call this function to all  iso under the device and returns them. Then in the main function add these iso to the menu

cmds/webboot/types.go
```diff
@@ -1,25 +1,34 @@
...
-	},
-	&BookMarkISO{
-		url:   wbtcpURL,
-		label: "Webboot Tinycorepure",
-		name:  "TinyCorePure64.iso",
-	},
-}
+var (
+	bookmark = map[string]([]*BookMarkISO){
+		"bookmark Tinycore1": []*BookMarkISO{
```
pallaud, line 16: I think since the map is already called bookmark you don't need to have 'bookmark' in the keys here

cmds/webboot/webboot.go
```diff
@@ -108,3 +149,50 @@ func download(URL, fPath string) error {
 	}
 	return nil
 }
+
+// will call a function to get all the block devices, then mount it one by one
```
rminnich, line 153: For consistency with your other good comments, :
// getCachedIsos will call a function ...

cmds/webboot/webboot_test.go
```diff
@@ -105,3 +109,132 @@ func TestDownloadByBookmarkOption(t *testing.T) {
...
+			label: "test cached iso3",
+			group: "test group 2",
+		},
+		&CachedISO{
+			label: "test cached iso4",
+			group: "test group 2",
+		},
+	}
+
+	for _, tt := range []struct {
```
rminnich, line 148: This is really neat. This it the first time, EVERY, in this project, and its predecessor, that we've ever had UI testing. Very nice :-)

cmds/webboot/webboot.go
```diff
@@ -66,6 +78,36 @@ func (d *DownloadByLink) Exec() error {
 	return nil
 }
 
+// Exec let user to choose a group of the cached iso
+func (i *InstallCachedISO) Exec() error {
+	isos := i.cachedISO
+	groupedIso := make(map[string]([]*CachedISO))
```
pallaud, line 84: I wonder if having the 'group' be a field of a struct and having to iterate through all ISOs to organize by group every time this menu is presented is a little inefficient. Is there a way we can do this in levels instead? So I'm imaging a directory structure for the cached ISOs that looks something like this

```
TinyCore/
   TinyCorePure64.iso
Ubuntu/
   Ubuntu-20.04.iso
   Ubuntu-18.04.iso
ArchLinux/
```
So depending on how you've worked out the directory structure expected on the USB device, would it make sense to have the menu display one level of directories at a time? Then once one is selected, the Exec for that can find subdirectories / isos in the given directory and show the next menu? That way you're not having to do all this classification every time. I think this might also simplify your types.

@rminnich any thoughts on this?

cmds/webboot/webboot.go
xsy1996: sorry it is not what getCachedIsos will do. it's just a todo to remind me what to do next. I will mark them by "todo:" . sorry

cmds/webboot/webboot.go
xsy1996: I think that's a good idea! thank you !

cmds/webboot/webboot.go
```diff
@@ -66,11 +78,50 @@ func (d *DownloadByLink) Exec() error {
 	return nil
 }
 
+// Exec displays subdirectory or cached isos under the path directory
+func (g *DirGroup) Exec() error {
+
```
pallaud, line 83: nit: remove space

cmds/webboot/types.go
```diff
@@ -38,29 +45,61 @@ func (b *BookMarkISO) Label() string {
...
+// Label is the string this option displays in the menu page.
 func (d *DownloadByLink) Label() string {
 	return "Download by link"
 }
+
+// DirGroup option displays subdirectory or cached isos under a certain directory
+// InstallCachedISO option is a special DirGroup option
+// which represents the root of the cache directory
+type DirGroup struct {
+	name     string
```
pallaud, line 85: For consistency maybe call this label instead of name?

cmds/webboot/webboot.go
```diff
@@ -66,11 +78,50 @@ func (d *DownloadByLink) Exec() error {
...
+	entries := []menu.Entry{}
+
+	readerInfos, err := ioutil.ReadDir(g.path)
+	if err != nil {
+		return err
+	}
+
+	// check the directory, if there is a subdirectory, add another DirGroup option
+	// if there is iso file, add a chacedIso option
```
pallaud, line 92: nit: spelling of 'cached' here and below

cmds/webboot/webboot.go
```diff
@@ -66,11 +78,50 @@ func (d *DownloadByLink) Exec() error {
...
+			})
+		} else if filepath.Ext(info.Name()) == ".iso" {
+			iso := &CachedISO{
+				path:  filepath.Join(g.path, info.Name()),
+				label: info.Name(),
+			}
+			entries = append(entries, iso)
+		}
+	}
+	_, err = menu.DisplayMenu("Groups", "Choose a group of chached iso", entries, g.uiEvents)
```
pallaud, line 108: I think the menu could just be titled something like "Distros" and just say something like "Choose an option:" - I don't think the group terminology is clear

cmds/webboot/webboot.go
```diff
@@ -66,11 +78,50 @@ func (d *DownloadByLink) Exec() error {
...
+	return err
+}
+
+// linkOpen sets up a http session to the URL
 func linkOpen(URL string) (io.ReadCloser, error) {
 	u, err := url.Parse(URL)
 	if err != nil {
 		log.Fatal(err)
 	}
+	// if the link is not begin with "file", "https" and "http", we do not support it
```
pallaud, line 124: nit: I think this comment actually isnt needed because by looking at the error message on link 138, someone would understand this already

cmds/webboot/webboot.go
```diff
@@ -108,3 +158,28 @@ func download(URL, fPath string) error {
 	}
 	return nil
 }
+
+// getHierachyMenu makes a hierarchy menu:
+// level 1: 0.Install cached ISO;1.Download ISO by bookmark; 2.Download ISO by link
```
pallaud, line 163: I think in this comment its enough to just say what the 3 options are, and mention that the menu will display nested levels of directories until an ISO is chosen, or something along those lines. Lets try to keep the comments short if we can. The specifics of each option can be talked about in the comments for those specific functions.

cmds/webboot/webboot_test.go
```diff
@@ -105,3 +109,98 @@ func TestDownloadByBookmarkOption(t *testing.T) {
...
+		{
+			name:      "hit second iso in the second dir",
+			userInput: []string{"1", "<Enter>", "1", "<Enter>"},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			uiEvents := make(chan ui.Event)
+			go pressKey(uiEvents, tt.userInput)
+
+			var entry menu.Entry = &DirGroup{uiEvents: uiEvents, path: "./testdata/", name: "test install cached iso option"}
```
pallaud, line 131: Since this is just a test, you don't need to provide a name in this struct, it'll default to ""

cmds/webboot/webboot.go
```diff
@@ -24,23 +26,33 @@ func (b *BookMarkISO) Exec() error {
...
+// Exec let user to choose a group of the bookmark
+func (d *DownloadByBookmark) Exec() error {
+	groups := []menu.Entry{}
+	for key, value := range bookmarks {
+		entries := []menu.Entry{}
+		for _, bm := range value {
+			bm.uiEvents = d.uiEvents
+			entries = append(entries, bm)
+		}
+		groups = append(groups, &BookmarkGroup{
```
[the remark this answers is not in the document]
xsy1996: you mean every level of the bookmark is a list of entries. and if it has many level, the upper level entries will contain another list of entries inside?

cmds/webboot/webboot.go
```diff
@@ -66,6 +59,52 @@ func (d *DownloadByLink) Exec() error {
...
+}
+
+// Exec displays the next level of bookmarks
+func (g *BookmarkGroup) Exec() error {
+	nextLevel := []menu.Entry{}
+	for _, e := range g.entries {
+		if b, ok := e.(*BookmarkGroup); ok {
+			b.uiEvents = g.uiEvents
+			nextLevel = append(nextLevel, b)
+		} else {
```
xsy1996, line 98: I think there is some redundancy but I cannot simplify it. In every BookmarkGroup option, I have to re-traverse all entries and add uiEvents to them one by one. So the benefit of direct hardcode bookmarks is reduced, I still need to reprocess entries instead of using them directly. But besides that, I have no other idea to add uiEvents to the entries in the next level. @pallaud

xsy1996:
Screenshot from 2020-07-30 13-03-23
Screenshot from 2020-07-30 13-03-32
Screenshot from 2020-07-30 13-03-43
Screenshot from 2020-07-30 13-04-13
Screenshot from 2020-07-30 13-04-19
Screenshot from 2020-07-30 13-04-24
Screenshot from 2020-07-30 13-04-32
Screenshot from 2020-07-30 13-04-55
Screenshot from 2020-07-30 13-05-32
Screenshot from 2020-07-30 13-06-08
Screenshot from 2020-07-30 13-06-16
Screenshot from 2020-07-30 13-06-25

cmds/webboot/webboot.go
```diff
@@ -66,6 +59,52 @@ func (d *DownloadByLink) Exec() error {
...
+			nextLevel = append(nextLevel, e)
+		}
+
+	}
+	_, err := menu.DisplayMenu("Bookmarks", "Choose an option", nextLevel, g.uiEvents)
```
pallaud, line 103: I'm thinking we might be able to avoid having to add all these ui.Events in the structs. The DisplayMenu function returns the entry that was chosen, and if you break up the Exec function to call a helper function that takes in a ui.Events and returns the result of the next menu, would that helper function then give you everything you need for a test?

cmds/webboot/webboot.go
pallaud: For the sake of making this PR a little less complicated, can we just have the bookmarks show up as a single list for now, and not have any hierarchy? In a follow up PR we can look into doing that but for now I think it'd be better to just see this work first.

cmds/webboot/types.go
```diff
@@ -1,32 +1,44 @@
...
+					},
+				},
+			},
+		},
+	}
+)
 
 // BookMarkISO contains information of the iso user want to download
 // expected to implement Entry interface
 type BookMarkISO struct {
```
pallaud, line 39: I'm wondering if this has to be a different struct from CachedISO; is there a way it doesn't have to be?

xsy1996: Yes, BookMarkISO's exec will download the iso first then boot and the cachedISO will just boot it.

PR #100, opened 2020-07-31
Update go.mod

xsy1996:
Let all imported internal pkg point to the local pkg file. So when the local pkg is modified, it can take effect immediately, without pushing to the remote repo first.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,3 +20,5 @@ require (
 	google.golang.org/grpc v1.27.1
 	pack.ag/tftp v1.0.0
 )
+
+replace github.com/u-root/webboot v0.0.0 => /gopath/github.com/u-root/webboot 


orangecms:
Is the mix of both `go.mod` and `Gopkg.toml` an issue? I couldn't make use of the updated dependencies really, tried various `go mod tidy` and `dep ensure -update` and `go get -u` commands to make use of newer features in u-root. It did work eventually when I bluntly removed `go.{mod,sum}` though, see https://github.com/orangecms/webboot/commits/local. Go is puzzling me and I cannot really find good hints on the web for some reason. It would be good if we got that figured out. I did notice that u-root itself doesn't use the new Go modules yet, which is what created `go.mod` iiuc.

PR #103, opened 2020-07-31
Add func main() & Remove Entry.Exec()

xsy1996:
1. add main func which makes a menu that only contains download option. this option allow user to download from bookmark or use custom url.
2. removed the Exec from Entry interface. because the Exec in the Entry interface is hard to test. Now when an entry is choose, main func calls it's execute logic and get the chosen entry in the next level.  Then do this repeatly until  there is not next level of option.
3. simplified the menus. Combines the DownloadByLink and DownloadByBookmark options to Download option. If the iso user want is already recored in bookmark, use it, elsewise ask for a download link.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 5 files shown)


cmds/webboot/webboot.go
```diff
@@ -10,60 +11,56 @@ import (
...
+		if link, err = menu.NewInputWindow("Enter URL:", menu.AlwaysValid, uiEvents); err != nil {
+			return nil, err
+		}
 	}
-	// todo: boot the iso
-	log.Printf("ISO is downloaded at %s", fPath)
-	return nil
+
+	fpath := filepath.Join("/tmp", filename)
+	if err = download(link, fpath); err != nil {
```
pallaud, line 54: I forgot did we ever end up adding a progress bar for the download?

cmds/webboot/webboot_test.go
```diff
@@ -44,64 +43,76 @@ func TestDownload(t *testing.T) {
...
+
+			downloadOption := DownloadOption{}
+			entry, err := downloadOption.exec(uiEvents)
+
+			if err != nil {
+				t.Errorf("Error: %v", err)
+			}
+			iso, ok := entry.(*ISO)
+			if !ok {
+				t.Errorf("Fail to transfor the entry to *ISO, error msg: %+v", err)
```
pallaud, line 88: I think here the error would be "expected type ISO"

cmds/webboot/webboot_test.go
```diff
@@ -44,64 +43,76 @@ func TestDownload(t *testing.T) {
...
+			}
+			if tt.want.label != iso.label || tt.want.path != iso.path {
+				t.Errorf("Incorrect return. get %+v, want %+v", entry, tt.want)
+			}
+			if _, err := os.Stat(iso.path); err != nil {
+				t.Errorf("Fail to find downloaded file, error msg: %+v", err)
+			}
+
+			if err := os.Remove(iso.path); err != nil {
+				t.Errorf("Fail to remove test file, error msg: %+v", err)
```
pallaud, line 98: For future reference, I think its ok if you dont say "error msg:", you can just put a colon after your message and print the err

cmds/webboot/webboot.go
xsy1996: we haven't add this. but I will do some research about how to add it now.

PR #104, opened 2020-08-01
Complete the menu

xsy1996:
Add "Use Cached ISO" opton to main menu. which will displays the cache directory nested, until reach the iso file. change the structure of cache directory will change the hierarch of menu at the same time.

### DIFF
--- a/cmds/webboot/Readme.md
+++ b/cmds/webboot/Readme.md
@@ -6,3 +6,9 @@ The webboot program would do the following:
 
 ### Test
 Our UI uses a package called Termui. Termui will parse the standard input into keyboard events and insert them into a channel, then from which the Termui get it's input.  For implement a unattended test, I manually build a series of keyboard events that reperesent my intented input for test, and insert them into a channel. Then I replace the original input channel with my channel in the test. So the go test could run a test of ui automatically.
+
+See TestDownloadOption for an example:
+ - create a channel by make(chan ui.Event).
+ - use go pressKey(uiEvents, input) to translate the intented test input to keyboard events and push them to the uiEvents chanel.
+ - use the uiEvents channel by call downloadOption.exec(uiEvents). (Main function will always call ui.PollEvents() to get the sandard input channel) 
+ - all functions involving in ui input will provide a argument to indicate the input chanel.
--- a/cmds/webboot/Readme.md
+++ b/cmds/webboot/Readme.md
@@ -1,0 +1,8 @@
+### Usage
+The webboot program would do the following:
+ - Present a menu with the existing cached distro options
+ - If the user wants a distro that is not cached, they can download an ISO 
+ - After the user decides on an ISO, boot it.
+
+### Test
+Our UI uses a package called Termui. Termui will parse the standard input into keyboard events and insert them into a channel, then from which the Termui get it's input.  For implement a unattended test, I manually build a series of keyboard events that reperesent my intented input for test, and insert them into a channel. Then I replace the original input channel with my channel in the test. So the go test could run a test of ui automatically.
--- a/cmds/webboot/utils.go
+++ b/cmds/webboot/utils.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+)
+
+func linkOpen(URL string) (io.ReadCloser, error) {
+	u, err := url.Parse(URL)
+	if err != nil {
+		log.Fatal(err)
+	}
+	switch u.Scheme {
+	case "file":
+		return os.Open(URL[7:])
+	case "http", "https":
+		resp, err := http.Get(URL)
+		if err != nil {
+			return nil, err
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("HTTP Get failed: %v", resp.StatusCode)
+		}
+		return resp.Body, nil
+	}
+	return nil, fmt.Errorf("%q: linkopen only supports file://, https://, and http:// schemes", URL)
+}
+
+// download will download a file from URL and save it as fPath
+// todo: add a download progress bar
+func download(URL, fPath string) error {
+	isoReader, err := linkOpen(URL)
+	if err != nil {
+		return err
+	}
+	defer isoReader.Close()
+	f, err := os.Create(fPath)
+	if err != nil {
+		return err
+	}
+	if _, err := io.Copy(f, isoReader); err != nil {
+		return fmt.Errorf("Fail to copy iso to a persistent memory device: %v", err)
+	}
+	if err = f.Close(); err != nil {
+		return fmt.Errorf("Fail to  close %s: %v", fPath, err)
+	}
+	verbose("%q is downloaded at %q\n", URL, fPath)
+	return nil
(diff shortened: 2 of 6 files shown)


xsy1996:
This is a screen recording of test
cinnamon-20200731-3

rminnich:
This is great. It is something I don't think I'd ever figure out in a million years. Wow.

rminnich:
I did run it and enter ab.dc for the URL and the program exited, it would be nice if it was a bit more gracious.

Also do I have a way to see what bookmarks there are.

xsy1996:
Sorry I think the reason enter ab.dc for the URL and the program exited is because I mistakely only print the error message  in the verbose mode. Actually the previous version does have a way to see what bookmarks there are. But the menu may looks complicated. So Prachi suggesst to just let user input the name of iso they want. And if we can find a matched name in the bookmark, we use the book mark, elsewise we ask for a download link. The biggest problem of menu is  termui's ui is not scrollable so the length of menu is limited. so if we want to display the bookmark, we may have to split the bookmark into small pieces and display them level by level

xsy1996:
demo of entering invalid download link
wrong_download_link

cmds/webboot/webboot.go
```diff
@@ -51,69 +47,60 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
 	fpath := filepath.Join("/tmp", filename)
-	if err = download(link, fpath); err != nil {
-		return nil, err
-	}
-
-	var iso *ISO = &ISO{
-		label: filename,
-		path:  fpath,
+	// if download link is not valid, ask again until the link is rights
+	for err = download(link, fpath); err != nil; err = download(link, fpath) {
```
pallaud, line 51: Why do we need to do `err = download(link, fpath)` twice? Hm would it work to do the loop this way

```
derr := download(link, fpath)
for derr != nil {
  // display derr
 // get input
  derr = download(link, fpath)
}
```

cmds/webboot/webboot.go
```diff
@@ -108,12 +132,16 @@ func download(URL, fPath string) error {
...
-	log.Println(*v)
 	if *v {
 		verbose = log.Printf
 	}
 
 	entries := []menu.Entry{
+		&DirOption{
+			label: "Use Cached ISO",
+			// todo: replace ./testdata with cache directory
+			path: "./testdata",
```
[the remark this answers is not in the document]
pallaud: Can we do this in this PR itself? It shouldn't be too hard to do so its preferable to do it now and not hardcode testdata.

pkg/menu/menu.go
```diff
@@ -84,15 +84,20 @@ func processInput(introwords string, location int, wid int, ht int, isValid vali
 				warning.Text = ""
 				ui.Render(warning)
 			}
-			input.Text += k
+			if k == "<Space>" {
+				input.Text += " "
+			} else if k[0:1] != "<" {
```
pallaud, line 89: I'm a little confused, why do we need this check?

xsy1996: because for example we hit left arrow, the termui's ui.PollEvents() will push a string"<Left>" to the channal. and all other special key like arrow, F1,etc on the keyboard will be treat in this way. So if the key is not begin at '<' it means it's a alphabet or number or .

cmds/webboot/webboot.go
xsy1996: Ok that would works too thanks. the reason I write download(link, fpath) twice here because the for loop's first statement is a initialization, second is the judgement; third is the statement will run at the begin of every loop.

cmds/webboot/webboot.go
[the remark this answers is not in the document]
xsy1996: sorry the revise of PR haven't done yet, so I do not call you to review. I plan to fix this part in the next commit

cmds/webboot/webboot_test.go
```diff
@@ -104,15 +105,46 @@ func TestDownloadOption(t *testing.T) {
...
+				t.Errorf("Fail to execute option (%q)'s exec(): %+v", entry.Label(), err)
+				break
+			}
+		} else if iso, ok := entry.(*ISO); ok {
+			if iso.label != wanted.label || iso.path != wanted.path {
+				t.Errorf("Get wrong chosen iso. get %+v, want %+v", iso, wanted)
+			}
+			break
+		} else {
+			t.Errorf("Unknow type, get a %T of entry %+v", entry, entry)
```
pallaud, line 146: Nit: `t.Errorf("Unknown type. got entry of type %T, wanted DirOption or ISO", entry)`

cmds/webboot/utils.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+)
+
+// WriteCounter counts the number of bytes written to it. It implements to the io.Writer
```
pallaud, line 13: Could you move the progress bar change to another PR? Lets try and keep each PR focused on 1 thing.

cmds/webboot/Readme.md
```diff
@@ -0,0 +1,8 @@
+### Usage
+The webboot program would do the following:
+ - Present a menu with the existing cached distro options
+ - If the user wants a distro that is not cached, they can download an ISO 
+ - After the user decides on an ISO, boot it.
+
+### Test
+Our UI uses a package called Termui. Termui will parse the standard input into keyboard events and insert them into a channel, then from which the Termui get it's input.  For implement a unattended test, I manually build a series of keyboard events that reperesent my intented input for test, and insert them into a channel. Then I replace the original input channel with my channel in the test. So the go test could run a test of ui automatically.
```
pallaud, line 8: Could you give an example here maybe? Say something like "See TestDownloadOption for an example. We use `go pressKey(uiEvents, input)` to set up the channel..."

cmds/webboot/webboot.go
```diff
@@ -51,69 +48,65 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
 func main() {
 	flag.Parse()
-	log.Println(*v)
 	if *v {
 		verbose = log.Printf
 	}
 
+	cachedDir := *dir
+	if cachedDir == "" {
+		// todo: find the cache directory by check block devices
```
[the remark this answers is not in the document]
xsy1996: Oh yeah, that would be better thanks.

pkg/menu/menu.go
```diff
@@ -84,15 +84,20 @@ func processInput(introwords string, location int, wid int, ht int, isValid vali
 				warning.Text = ""
 				ui.Render(warning)
 			}
-			input.Text += k
+			if k == "<Space>" {
```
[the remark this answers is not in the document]
xsy1996: I think <Space> is just a special input character. if i make it another switch statement, I should repeat the logic like checking warning text etc.

cmds/webboot/utils.go
xsy1996: Ok sure

pkg/menu/menu.go
xsy1996: ok, will do

PR #106, opened 2020-08-05
Add a download progress bar

xsy1996:
Currently the progress bar only print out how many bytes have been downloaded,  it cannot show how many percentages have been downloaded. Because we do not know what the whole size of the file. The io.Copy() just read the online file segment by segment, it do not know how big the file is until it reach the end of file.


cmds/webboot/network_utils.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (wc *WriteCounter) Write(p []byte) (int, error) {
+	n := len(p)
+	wc.Total += uint64(n)
+	wc.PrintProgress()
+	return n, nil
+}
+
+// PrintProgress will print how many bytes have been writen to the file
+func (wc WriteCounter) PrintProgress() {
```
pallaud, line 26: Why have a separate function for this, does it work to just put these 2 lines directly in the Write function?

xsy1996: I think that putting them in a separate function will make the logic clearer, because I think Write() is not neccesarily equal to PrintProgress(). For example, it is possible that we will do other things in Write in addition to the progress bar. Although I can't think of what else needs to do now, I think separating them may be more flexible.

cmds/webboot/utils.go
```diff
@@ -7,8 +7,23 @@ import (
...
+type WriteCounter struct {
+	Total uint64
+}
+
+func (wc *WriteCounter) Write(p []byte) (int, error) {
+	n := len(p)
+	wc.Total += uint64(n)
+	// print how many bytes have been writen to the file
+	fmt.Printf("\r%s", strings.Repeat(" ", 40))
+	fmt.Printf("\rDownloading... %v bytes complete", wc.Total)
```
pallaud, line 23: How often will this message print?

xsy1996: It depends on how offten io.Copy() read a segment of data from the online file. I think it may involves  the speed of internet. In the test the message update many times per second.

xsy1996: since we use /r here, the new message just overwrite the old message instead of create a new line. if that is what your worry about

pallaud: Ok - I'll approve this but would you mind also including a quick screengrab of what it looks like here for future reference?

xsy1996: ok will do

xsy1996:
the test of progress bar
test-progressbar

PR #107, opened 2020-08-05
Add getCachedDirectory

xsy1996:
The main function will go find the cache directory in the usb stick if the -dir argument is empty.


cmds/webboot/utils.go
```diff
@@ -66,3 +71,28 @@ func download(URL, fPath string) error {
...
+		log.Fatal("No available block devices to boot from")
+	}
+
+	mountPoints, err := ioutil.TempDir("", "temp-device-")
+	if err != nil {
+		return nil, fmt.Errorf("cannot create tmpdir: %v", err)
+	}
+
+	for _, device := range blockDevs {
+		mp, err := mount.TryMount("/dev/"+device.Name, filepath.Join(mountPoints, device.Name), "", mount.ReadOnly)
```
rminnich, line 89: filepath.Join("/dev", device.Name)

cmds/webboot/webboot.go
```diff
@@ -97,9 +97,16 @@ func main() {
...
 	cachedDir := *dir
+
 	if cachedDir == "" {
-		// todo: find the cache directory by check block devices
-		log.Fatal("Please provide a valid dir argument")
+		mp, err := getCachedDirectory()
+		if err != nil {
+			log.Fatalf("Fail to find the USB stick: %+v", err)
+		}
+		if mp == nil {
```
rminnich, line 106: getCachedDirectory should always return nil if there is an error.

cmds/webboot/webboot_test.go
```diff
@@ -145,3 +146,35 @@ func TestDirOption(t *testing.T) {
...
+// |  +-- tinycore2
+// |     +-- TinyCorePure64-10.1.iso
+// |     +-- TinyCorePure64-10.1 (copy).iso
+// ...
+func TestGetCachedDirectory(t *testing.T) {
+	mp, err := getCachedDirectory()
+	if err != nil {
+		t.Fatalf("Fail to find the USB stick: %+v", err)
+	}
+	if mp == nil {
```
rminnich, line 166: getCachedDirectory should always return nil if err is != nil?

xsy1996: Yes, the function returns both mount point  and error. and if there is a error, function always return nil for the mount point.

cmds/webboot/webboot.go
xsy1996: do you mean I should not log.Fatal in the getCachedDirectory function?

cmds/webboot/webboot_test.go
xsy1996: Oh I think I got you do you mean I should not check mp == nil here because if there is a error the mp will be nil too? I think if there is a error the t.Fatalf("Fail to find the USB stick: %+v", err) will be call and the test will exit immediately? but I will fix that thanks!

cmds/webboot/utils.go
```diff
@@ -66,3 +71,28 @@ func download(URL, fPath string) error {
 	verbose("%q is downloaded at %q\n", URL, fPath)
 	return nil
 }
+
+// getCachedDirectory recognizes the usb stick that contains the cached directory from block devices,
```
pallaud, line 75: I think either in this comment or in the README you'd want to explain the expected file structure on the USB stick so that someone using webboot would understand why this is the way it is.

cmds/webboot/webboot.go
```diff
@@ -97,9 +97,13 @@ func main() {
...
 	cachedDir := *dir
+
 	if cachedDir == "" {
-		// todo: find the cache directory by check block devices
-		log.Fatal("Please provide a valid dir argument")
+		mp, err := getCachedDirectory()
+		if err != nil {
+			log.Fatalf("Fail to find the USB stick: %+v", err)
+		}
+		cachedDir = filepath.Join(mp.Path, "Image")
```
pallaud, line 106: Was it "Image" or "Images"?

cmds/webboot/utils.go
```diff
@@ -66,3 +71,28 @@ func download(URL, fPath string) error {
...
+// and return it's mount point.
+func getCachedDirectory() (*mount.MountPoint, error) {
```
pallaud, line 77: This is a personal preference but this seems like something pretty specific to the weboot program and not necessarily a util, so I'd put it in webboot.go

cmds/webboot/utils.go
```diff
@@ -66,3 +71,28 @@ func download(URL, fPath string) error {
...
+	for _, device := range blockDevs {
+		mp, err := mount.TryMount(filepath.Join("/dev/", device.Name), filepath.Join(mountPoints, device.Name), "", mount.ReadOnly)
+		if err != nil {
+			continue
+		}
+		if _, err = os.Stat(filepath.Join(mp.Path, "Image")); err == nil {
+			return mp, nil
+		}
+	}
+	return nil, fmt.Errorf("Do not find the cache directory")
```
pallaud, line 97: I think in this error message you'd similarly want to explain to the user what was expected - you expected to find a directory named Images/ at the root of the block device right?

cmds/webboot/webboot_test.go
```diff
@@ -140,8 +141,38 @@ func TestDirOption(t *testing.T) {
...
+// which structure is
+// +-- Image
+// |  +-- tinycore1
+// |  |  +-- TinyCorePure64.iso
+// |  |  +-- TinyCorePure64 (copy).iso
+// |  +-- tinycore2
+// |     +-- TinyCorePure64-10.1.iso
+// |     +-- TinyCorePure64-10.1 (copy).iso
+// ...
+func TestGetCachedDirectory(t *testing.T) {
```
pallaud, line 162: If this test doesn't always pass I'm not sure that we want to include it with the rest of the unit tests...I think with our CI we'd want to run all the unit tests to confirm that they all pass, and if this one always fails that might throw it off? I'm not totally sure how CIs work so maybe this isn't a problem @rminnich

xsy1996: I deleted them for now

PR #109, opened 2020-08-06
Complete Boot Process 

xsy1996:
able to choose a iso from menu and boot it now.


xsy1996:
Test of using cachec iso. It could find the cached directory in the USB successfully.
testCachedISO

xsy1996:
Test of using download way.
testDownloadISO

cmds/webboot/webboot.go
```diff
@@ -8,20 +8,40 @@ import (
...
 	"github.com/u-root/webboot/pkg/menu"
 )
 
 var (
 	v       = flag.Bool("verbose", false, "Verbose output")
 	verbose = func(string, ...interface{}) {}
 	dir     = flag.String("dir", "", "Path of cached directory")
-	network = flag.Bool("network", true, "Should  or not set up network")
+	network = flag.Bool("network", true, "Should or not set up network")
+	boot    = flag.Bool("boot", true, "Should or not boot the iso. May trun it off for test.")
```
pallaud, line 20: Usually we name such flags dry-run, so if dry-run is true then we wouldn't boot the iso

cmds/webboot/webboot.go
```diff
@@ -8,20 +8,40 @@ import (
...
-	// todo: boot the iso
-	log.Printf("ISO is at %s\n", i.path)
+func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
+	verbose("Intent to boot %s", i.path)
+	configs, err := bootiso.ParseConfigFromISO(i.path)
+	if err != nil {
+		return err
+	}
+	verbose("Get configs: %+v", configs)
+	if boot {
```
pallaud, line 31: It would be better to have 

```
if !boot {
    return nil
}
// rest of the logic
```

cmds/webboot/webboot.go
```diff
@@ -8,20 +8,40 @@ import (
...
+		entries := []menu.Entry{}
+		for _, config := range configs {
+			entries = append(entries, &Config{label: config.Label()})
+		}
+		if c, err := menu.DisplayMenu("Configs", "Choose an option", entries, uiEvents); err == nil {
```
pallaud, line 36: I think it might be a little cleaner to do

```
c, err := menu.DisplayMenu("Configs", "Choose an option", entries, uiEvents)
if err == nil {
    err = bootiso.BootFromPmem(i.path, c.Label())
}
return err
```

xsy1996: That's so concise and beautiful!

PR #112, opened 2020-08-07
Still show a menu if the cache dir is not found

xsy1996:
It will return a menu only include download option if the cache dir is not found

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -128,23 +128,23 @@ func main() {
 		verbose = log.Printf
 	}
 
-	cachedDir := *dir
-
-	if cachedDir == "" {
-		mp, err := getCachedDirectory()
-		if err != nil {
-			log.Fatalf("Fail to find the USB stick: %+v", err)
-		}
-		cachedDir = filepath.Join(mp.Path, "Image")
-	}
 	entries := []menu.Entry{
-		// "Use Cached ISO" option is a special DirGroup Entry
-		// which represents the root of the cache directory
-		&DirOption{
-			label: "Use Cached ISO",
-			path:  cachedDir,
-		},
 		&DownloadOption{},
+	}
+
+	if *dir == "" {
+		mp, err := getCachedDirectory()
+		if err == nil {
+			entries = append(entries, &DirOption{
+				label: "Use Cached ISO",
+				path:  filepath.Join(mp.Path, "Image"),
+			})
+		}
+	} else {
+		entries = append(entries, &DirOption{
+			label: "Use Cached ISO",
+			path:  *dir,
+		})
 	}
 
 	entry, err := menu.DisplayMenu("Webboot", "Choose an ISO:", entries, ui.PollEvents())


xsy1996:
test
cinnamon-20200807-11

cmds/webboot/webboot.go
```diff
@@ -128,26 +149,14 @@ func main() {
 		verbose = log.Printf
 	}
 
-	cachedDir := *dir
-
-	if cachedDir == "" {
-		mp, err := getCachedDirectory()
-		if err != nil {
-			log.Fatalf("Fail to find the USB stick: %+v", err)
```
pallaud, line 136: Looks like you've made some good progress towards figuring out the back option here - I hate to bring this up once you've already done this work, but I had imagined that this could be simpler for the user if you just, instead of having a log.Fatalf when you can't find a cache, just create entries without the "Use Cached ISO" option. So basically only like a 2-3 line change.

Then later once we have 1) the ability to cache a downloaded ISO and 2) a back option, we can revisit this and make it so that people can download an ISO, cache it, and then come back to the main menu and see this "Use Cached ISO" option.

Does that make sense? I think it will make things simpler.

xsy1996: That's a good idea, thank you!

xsy1996: Sorry I am i little bit not understand, because I have replaced the log.Fatal by a menu with only download option? Is that what you want?

xsy1996: ok I think I get you, but the process I imaged is , after the back option is ready, there will always be two option in the main menu, and if the cache dir is exist, the next menu contains the folders and files under the cache dir plus a back option,  and if the cache dir in not exist, the next menu only contains a back option.

cmds/webboot/webboot.go
```diff
@@ -129,22 +129,24 @@ func main() {
...
 	if cachedDir == "" {
 		mp, err := getCachedDirectory()
-		if err != nil {
-			log.Fatalf("Fail to find the USB stick: %+v", err)
+		if err == nil {
+			cachedDir = filepath.Join(mp.Path, "Image")
 		}
-		cachedDir = filepath.Join(mp.Path, "Image")
 	}
 	entries := []menu.Entry{
```
pallaud, line 138: What if you created entries before checking cachedDir, and then if cachedDir exists, append the option to entries - this way you don't have to check it twice

xsy1996: the problem is, the cached dir not only could be the cache dir in the USB, but also could be a local dir given by args, so if I insert entry here, I should also make another if conditional sentance that is when dir args is not empty. It is not so simple too

cmds/webboot/webboot.go
```diff
@@ -128,23 +128,23 @@ func main() {
...
+	entries := []menu.Entry{
+		&DownloadOption{},
+	}
 
-	if cachedDir == "" {
+	if *dir == "" {
 		mp, err := getCachedDirectory()
-		if err != nil {
-			log.Fatalf("Fail to find the USB stick: %+v", err)
+		if err == nil {
```
pallaud, line 137: Might be good to log the error here so that if someone was expecting to see a cached directory but doesn't, they can see what went wrong

xsy1996: ok, will revise it in the next PR.

PR #113, opened 2020-08-10
Add back option

xsy1996:
add back option & clean the code

### DIFF
--- a/cmds/webboot/Readme.md
+++ b/cmds/webboot/Readme.md
@@ -12,3 +12,10 @@ See TestDownloadOption for an example:
  - use go pressKey(uiEvents, input) to translate the intented test input to keyboard events and push them to the uiEvents chanel.
  - use the uiEvents channel by call downloadOption.exec(uiEvents). (Main function will always call ui.PollEvents() to get the sandard input channel) 
  - all functions involving in ui input will provide a argument to indicate the input chanel.
+
+ ### Hint
+ If want to set up a cached directory in side the USB stick, the file structure of USB stick should be
++-- USB root
+|  +-- Images (<--- the cache directory. It must be named as "Images")
+|     +-- subdirectories or iso files
+...
--- a/cmds/webboot/types.go
+++ b/cmds/webboot/types.go
@@ -43,6 +43,16 @@ func (d *DownloadOption) Label() string {
 	return "Download an ISO"
 }
 
+// UseCacheOption let user boot a cached iso
+// expected to implement Entry interface
+type UseCacheOption struct {
+}
+
+// Label is the string this iso displays in the menu page.
+func (u *UseCacheOption) Label() string {
+	return "Use Cached ISO"
+}
+
 // DirOption represents a directory under cache directory
 // DirOption options displays it's sub-directory or iso files
 type DirOption struct {
@@ -53,4 +63,14 @@ type DirOption struct {
 // Label is the string this option displays in the menu page.
 func (d *DirOption) Label() string {
 	return d.label
+}
+
+// BackOption let user back to the upper menu
+// expected to implement Entry interface
+type BackOption struct {
+}
+
+// Label is the string this iso displays in the menu page.
+func (b *BackOption) Label() string {
+	return "Go Back"
 }
--- a/cmds/webboot/types.go
+++ b/cmds/webboot/types.go
@@ -1,5 +1,7 @@
 package main
 
+import "github.com/u-root/webboot/pkg/menu"
+
 const (
 	tcURL    = "http://tinycorelinux.net/10.x/x86_64/release/TinyCorePure64-10.1.iso"
 	wbtcpURL = "https://github.com/u-root/webboot-distro/raw/master/iso/tinycore/10.x/x86_64/release/TinyCorePure64.iso"
@@ -17,8 +19,9 @@ type ISO struct {
 	path  string
 }
 
+var _ = menu.Entry(&ISO{})
+
 // Label is the string this iso displays in the menu page.
-// expected to implement Entry interface
 func (i *ISO) Label() string {
 	return i.label
 }
@@ -28,37 +31,32 @@ type Config struct {
 	label string
 }
 
+var _ = menu.Entry(&Config{})
+
 // Label is the string this iso displays in the menu page.
 func (c *Config) Label() string {
 	return c.label
 }
 
 // DownloadOption let user download an iso then boot it
-// expected to implement Entry interface
 type DownloadOption struct {
 }
+
+var _ = menu.Entry(&DownloadOption{})
 
 // Label is the string this iso displays in the menu page.
 func (d *DownloadOption) Label() string {
 	return "Download an ISO"
 }
 
-// UseCacheOption let user boot a cached iso
-// expected to implement Entry interface
-type UseCacheOption struct {
-}
-
-// Label is the string this iso displays in the menu page.
-func (u *UseCacheOption) Label() string {
-	return "Use Cached ISO"
-}
-
 // DirOption represents a directory under cache directory
-// DirOption options displays it's sub-directory or iso files
+// it displays it's sub-directory or iso files
 type DirOption struct {
 	label string
 	path  string
 }
+
+var _ = menu.Entry(&DirOption{})
 
 // Label is the string this option displays in the menu page.
 func (d *DirOption) Label() string {
@@ -66,9 +64,10 @@ func (d *DirOption) Label() string {
 }
 
 // BackOption let user back to the upper menu
-// expected to implement Entry interface
 type BackOption struct {
 }
+
+var _ = menu.Entry(&BackOption{})
 
 // Label is the string this iso displays in the menu page.
 func (b *BackOption) Label() string {
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -24,8 +24,6 @@ var (
 	dryRun  = flag.Bool("dry_run", true, "If dry_run is true we won't boot the iso.")
 )
 
-var cacheDir = ""
-
 // ISO's exec downloads the iso and boot it.
 func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
 	verbose("Intent to boot %s", i.path)
@@ -173,18 +171,12 @@ func getMainMenu(cacheDir string) menu.Entry {
 	return entry
 }
 
-func goBackDir(currentPath string) menu.Entry {
-	backTo := filepath.Dir(currentPath)
-	entry := &DirOption{path: backTo}
-	return entry
-}
-
 func main() {
 	flag.Parse()
 	if *v {
 		verbose = log.Printf
 	}
-	cacheDir = *dir
+	cacheDir := *dir
 	if cacheDir != "" {
 		// call filepath.Clean to make sure the format of path is consistent
 		// we should check the cacheDir != "" before call filepath.Clean, because filepath.Clean("") = "."
@@ -225,7 +217,7 @@ func main() {
 					entry = getMainMenu(cacheDir)
 					break
 				}
-				entry = goBackDir(dirOption.path)
+				entry = &DirOption{path: filepath.Dir(dirOption.path)}
 			}
 		default:
 			log.Fatalf("Unknown type %T!\n", entry)
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -20,6 +20,8 @@ var (
 	boot    = flag.Bool("boot", true, "Should or not boot the iso. May trun it off for test.")
 )
 
+var cacheDir = ""
+
 // ISO's exec downloads the iso and boot it.
 func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
 	verbose("Intent to boot %s", i.path)
@@ -43,6 +45,23 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
 		}
 	}
 	return nil
+}
+
+// UseCacheOption's exec displays all possible cache directory
+// if the -dir args is valid show it in the menu
+// if it's able to find a cached diretory in the USB stick, show it too
+func (u *UseCacheOption) exec(uiEvents <-chan ui.Event, dir string) (menu.Entry, error) {
+	// filefath.Join an empty string to make sure the format of path is consistent
+	cacheDir = filepath.Join(dir, "")
+	if cacheDir == "" {
+		mp, err := getCachedDirectory()
+		if err != nil {
+			verbose("Fail to find the USB stick: %+v", err)
+			return menu.DisplayMenu("Webboot", "Choose an option:", []menu.Entry{&BackOption{}}, uiEvents)
+		}
+		cacheDir = filepath.Join(mp.Path, "Image")
+	}
+	return &DirOption{path: cacheDir}, nil
 }
 
 // DownloadOption's exec lets user input the name of the iso they want
@@ -97,7 +116,7 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool) (menu.Entr
 
 // DirOption's exec displays subdirectory or cached isos under the path directory
 func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
-	entries := []menu.Entry{}
+	entries := []menu.Entry{&BackOption{}}
 	readerInfos, err := ioutil.ReadDir(d.path)
 	if err != nil {
 		return nil, err
@@ -122,41 +141,38 @@ func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
 	return menu.DisplayMenu("Distros", "Choose an option", entries, uiEvents)
 }
 
+func getMainMenu() menu.Entry {
+	entries := []menu.Entry{
+		&UseCacheOption{},
+		&DownloadOption{},
+	}
+	entry, err := menu.DisplayMenu("Webboot", "Choose an option:", entries, ui.PollEvents())
+	if err != nil {
+		log.Fatal(err)
+	}
+	return entry
+}
+
 func main() {
 	flag.Parse()
 	if *v {
 		verbose = log.Printf
 	}
 
-	entries := []menu.Entry{
-		&DownloadOption{},
-	}
-
-	if *dir == "" {
-		mp, err := getCachedDirectory()
-		if err == nil {
-			entries = append(entries, &DirOption{
-				label: "Use Cached ISO",
-				path:  filepath.Join(mp.Path, "Image"),
-			})
-		}
-	} else {
-		entries = append(entries, &DirOption{
-			label: "Use Cached ISO",
-			path:  *dir,
-		})
-	}
-
-	entry, err := menu.DisplayMenu("Webboot", "Choose an ISO:", entries, ui.PollEvents())
-	if err != nil {
-		log.Fatal(err)
-	}
-
+	entry := getMainMenu()
+	var err error = nil
 	// check the chosen entry of each level
 	// and call it's exec() to get the next level's chosen entry.
 	// repeat this process until there is no next level
 	for entry != nil {
 		switch entry.(type) {
+		case *UseCacheOption:
+			if entry, err = entry.(*UseCacheOption).exec(ui.PollEvents(), *dir); err != nil {
+				log.Fatalf("Use Cached ISO option failed:%v", err)
+			}
+			if _, ok := entry.(*BackOption); ok {
+				entry = getMainMenu()
+			}
 		case *DownloadOption:
 			if entry, err = entry.(*DownloadOption).exec(ui.PollEvents(), *network); err != nil {
 				log.Fatalf("Download option failed:%v", err)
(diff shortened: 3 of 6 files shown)


xsy1996:
cinnamon-20200810-13

cmds/webboot/types.go
```diff
@@ -54,3 +64,13 @@ type DirOption struct {
...
+
+// BackOption let user back to the upper menu
+// expected to implement Entry interface
+type BackOption struct {
+}
+
+// Label is the string this iso displays in the menu page.
+func (b *BackOption) Label() string {
+	return "Go Back"
+}
```
rminnich, line 76: to ensure that things implement the interface you can do this
var _ = Entry(&BackOption{})

Then you don't need to say 'expected to implement' all the time, the code will make that clear.

cmds/webboot/webboot.go
```diff
@@ -16,10 +20,12 @@ var (
 	v       = flag.Bool("verbose", false, "Verbose output")
 	verbose = func(string, ...interface{}) {}
 	dir     = flag.String("dir", "", "Path of cached directory")
-	network = flag.Bool("network", true, "Should or not set up network")
-	boot    = flag.Bool("boot", true, "Should or not boot the iso. May trun it off for test.")
+	network = flag.Bool("network", true, "If dry-network is false we will not set up network")
+	dryRun  = flag.Bool("dry-run", true, "If dry-run is true we won't boot the iso. May trun it on for test.")
```
rminnich, line 24: what does trun mean? test run? or run?

cmds/webboot/webboot.go
```diff
@@ -28,21 +34,65 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
...
-	return nil
+	return nil, fmt.Errorf("Do not find the cache directory: Expected a /Images under at the root of a block device(USB)")
+}
+
+// UseCacheOption's exec displays all possible cache directory
+// if the -dir args is valid show it in the menu
+// if it's able to find a cached diretory in the USB stick, show it too
+func (u *UseCacheOption) exec(uiEvents <-chan ui.Event, dir string) (menu.Entry, error) {
+	// filefath.Join an empty string to make sure the format of path is consistent
+	cacheDir = filepath.Join(dir, "")
```
rminnich, line 86: filepath.Clean may be more useful here.
https://play.golang.org/p/IHPw1QDR85l
"Join joins any number of path elements into a single path, separating them with an OS specific Separator. Empty elements are ignored. The result is Cleaned. "

in other words Join calls Clean.

cmds/webboot/webboot.go
xsy1996: I think my expression is wrong. what I want to mean is, we may use this args in test, when we don't want to actually replace the kernel. I may  just say "If dry-run is true we won't boot the iso is ok."

cmds/webboot/types.go
xsy1996: Ok, thank you~

cmds/webboot/webboot.go
```diff
@@ -97,7 +147,7 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool) (menu.Entr
 
 // DirOption's exec displays subdirectory or cached isos under the path directory
 func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
-	entries := []menu.Entry{}
+	entries := []menu.Entry{&BackOption{}}
```
pallaud, line 150: This is just a personal preference but what do you think about having the back option be the last thing in the menu, not the first?

cmds/webboot/webboot.go
xsy1996: Thank you! That's helpful!

cmds/webboot/webboot_test.go
```diff
@@ -161,3 +161,33 @@ func TestGetCachedDirectory(t *testing.T) {
...
 }
+func TestBackOption(t *testing.T) {
+	uiEvents := make(chan ui.Event)
+	input := []string{"1", "<Enter>", "1", "<Enter>", "0", "<Enter>", "1", "<Enter>", "0", "<Enter>", "0", "<Enter>"}
+	go pressKey(uiEvents, input)
+
+	var entry menu.Entry = &DirOption{path: "./testdata"}
+	var currentPath string = ""
+	var err error = nil
+	for i := 0; i < 9; i++ {
```
pallaud, line 172: Hm I only looked at this briefly but why do we need this loop for 9 iterations? Is it possible to simplify this test to only 1) call exec for DirOption(/testdata), 2) choose 1 thing, 3) go back and confirm that the new option after going back has the correct path?

cmds/webboot/webboot.go
xsy1996: Ron I find there is a different between filepath.Clean and filepg.Join empty element, that is , when the raw path is empty too, the Clean will returns ".", while the Join will returns ""  Just for let you know.

cmds/webboot/Readme.md
```diff
@@ -11,4 +11,11 @@ See TestDownloadOption for an example:
...
  - use go pressKey(uiEvents, input) to translate the intented test input to keyboard events and push them to the uiEvents chanel.
  - use the uiEvents channel by call downloadOption.exec(uiEvents). (Main function will always call ui.PollEvents() to get the sandard input channel) 
- - all functions involving in ui input will provide a argument to indicate the input chanel.
\ No newline at end of file
+ - all functions involving in ui input will provide a argument to indicate the input chanel.
+
+ ### Hint
+ If want to set up a cached directory in side the USB stick, the file structure of USB stick should be
++-- USB root
+|  +-- Image (<--- the cache directory. It must be named as "Images")
```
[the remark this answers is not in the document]
xsy1996: Images, Sorry! I missed to change the name in the Readme.

cmds/webboot/webboot.go
```diff
@@ -16,10 +20,12 @@ var (
 	v       = flag.Bool("verbose", false, "Verbose output")
 	verbose = func(string, ...interface{}) {}
 	dir     = flag.String("dir", "", "Path of cached directory")
-	network = flag.Bool("network", true, "Should or not set up network")
-	boot    = flag.Bool("boot", true, "Should or not boot the iso. May trun it off for test.")
+	network = flag.Bool("network", true, "If dry-network is false we will not set up network")
```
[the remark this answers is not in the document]
xsy1996: network,sorry. I just changed it.

cmds/webboot/webboot.go
```diff
@@ -28,21 +34,65 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
...
 	}
-	return nil
+	return nil, fmt.Errorf("Do not find the cache directory: Expected a /Images under at the root of a block device(USB)")
+}
+
+// UseCacheOption's exec displays all possible cache directory
+// if the -dir args is valid show it in the menu
+// if it's able to find a cached diretory in the USB stick, show it too
+func (u *UseCacheOption) exec(uiEvents <-chan ui.Event, dir string) (menu.Entry, error) {
+	// filefath.Join an empty string to make sure the format of path is consistent
```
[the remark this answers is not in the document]
xsy1996: Sorry about that! I will be more careful

cmds/webboot/webboot.go
```diff
@@ -122,54 +172,61 @@ func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
 	for entry != nil {
 		switch entry.(type) {
+		case *UseCacheOption:
+			if entry, err = entry.(*UseCacheOption).exec(ui.PollEvents(), *dir); err != nil {
+				log.Fatalf("Use Cached ISO option failed:%v", err)
+			}
+			if _, ok := entry.(*BackOption); ok {
+				entry = getMainMenu()
+			}
 		case *DownloadOption:
```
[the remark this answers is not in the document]
xsy1996: Yeah I am thinking about this too. maybe if user input a special string like "back" in the input window, we go back?

cmds/webboot/webboot.go
xsy1996: ok, I will change that

cmds/webboot/webboot_test.go
xsy1996: I just think it’s easier to detect problems with a series of complicated instructions. But I agree with you now. Anyway, just keep it simple

cmds/webboot/webboot.go
```diff
@@ -28,21 +34,65 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
...
+// |  +-- Images (<--- the cache directory. It must be named as "Images")
+// |     +-- subdirectories or iso files
+// ...
+func getCachedDirectory() (*mount.MountPoint, error) {
+	blockDevs, err := block.GetBlockDevices()
+	if err != nil {
+		return nil, fmt.Errorf("No available block devices to boot from")
+	}
+
+	mountPoints, err := ioutil.TempDir("", "temp-device-")
```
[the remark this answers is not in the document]
xsy1996: Yeah I should not go to find the mount path every time because I already stored the cacheDir

cmds/webboot/webboot.go
```diff
@@ -122,54 +172,61 @@ func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
+	}
 
+	entry := getMainMenu()
+	var err error = nil
 	// check the chosen entry of each level
 	// and call it's exec() to get the next level's chosen entry.
 	// repeat this process until there is no next level
 	for entry != nil {
 		switch entry.(type) {
+		case *UseCacheOption:
```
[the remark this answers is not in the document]
xsy1996: Actually when I look at it I cannot remember the reason I have this separate struct at that time i wrote, But I remember there is a reason. Anyway, I'll change it and see if there is a problem.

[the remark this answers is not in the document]
xsy1996: I know what the reason is now. It is because in my image, if the cache dir is not exist, I will also show a option of "use cached ISO". and if it is hit, the menu only contains back option. And DirOption can not doing this because it cannot represent a path do not exist so I need UseCacheOption.  But when I think further, just do not show the "use cached ISO" option at all when i can not find the cache dir would be simpler. I'll change it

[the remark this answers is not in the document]
pallaud: Sounds good!

cmds/webboot/types.go
```diff
@@ -12,18 +12,19 @@ var bookmarks = map[string]string{
 }
 
 // ISO contains information of the iso user want to boot
+// var _ = Entry(&ISO{})
```
rminnich, line 15: Don't leave this commented out, right?  It's a variable declaration. It does not work if it is commented out.

xsy1996: oh so you mean just let it be part of the code. just define the var but not use? Interesting! I haven't imaged that it could be like this

cmds/webboot/types.go
```diff
@@ -17,8 +19,9 @@ type ISO struct {
 	path  string
 }
 
+var _ = menu.Entry(&ISO{})
```
pallaud, line 22: Sorry I might've missed the discussion but why are these necessary?

cmds/webboot/webboot.go
```diff
@@ -119,39 +123,83 @@ func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
+
+	entry, err := menu.DisplayMenu("Webboot", "Choose an option:", entries, ui.PollEvents())
 	if err != nil {
 		log.Fatal(err)
 	}
+	return entry
+}
+
+func goBackDir(currentPath string) menu.Entry {
+	backTo, _ := filepath.Split(currentPath)
```
pallaud, line 177: Does filepath.Dir not work for this?

cmds/webboot/webboot.go
```diff
@@ -16,10 +20,12 @@ var (
 	v       = flag.Bool("verbose", false, "Verbose output")
 	verbose = func(string, ...interface{}) {}
 	dir     = flag.String("dir", "", "Path of cached directory")
-	network = flag.Bool("network", true, "Should or not set up network")
-	boot    = flag.Bool("boot", true, "Should or not boot the iso. May trun it off for test.")
+	network = flag.Bool("network", true, "If network is false we will not set up network")
+	dryRun  = flag.Bool("dry_run", true, "If dry_run is true we won't boot the iso.")
 )
 
+var cacheDir = ""
```
pallaud, line 27: Is this necessary anymore? Seems like you pass a parameter cacheDir anyway

cmds/webboot/webboot.go
```diff
@@ -162,14 +210,24 @@ func main() {
...
-			if entry, err = entry.(*DirOption).exec(ui.PollEvents()); err != nil {
+			dirOption := entry.(*DirOption)
+			if entry, err = dirOption.exec(ui.PollEvents()); err != nil {
 				log.Fatalf("Directory option failed:%v", err)
 			}
+			if _, ok := entry.(*BackOption); ok {
+				// if dirOption.path == cacheDir means current dir is the root of cache dir
+				// and it should go back to the main menu.
+				if dirOption.path == cacheDir {
+					entry = getMainMenu(cacheDir)
```
pallaud, line 226: Hm so it seems like right now, every time the back option is chosen, getMainMenu will be called and display a menu that says "Weboot: choose an option" and contains the download option as well, right? But ideally 'back' just means go back to the menu page I just saw, not the main menu right?

Instead of having do to this, could you just set entry equal to that DirOption you're creating and leave it at that? Then you'd continue in the loop and display the same menu you would've displayed for the parent right? That takes out the need for creating this getMainMenu function and makes it so that users arent always seeing the main menu if they just want to go back.

Does that work?

cmds/webboot/webboot_test.go
```diff
@@ -133,31 +133,34 @@ func TestDirOption(t *testing.T) {
...
-		filepath.Join(mp.Path, "/Image/tinycore1/TinyCorePure64.iso"),
-		filepath.Join(mp.Path, "/Image/tinycore1/TinyCorePure64 (copy).iso"),
-		filepath.Join(mp.Path, "/Image/tinycore2/TinyCorePure64-10.1.iso"),
-		filepath.Join(mp.Path, "/Image/tinycore2/TinyCorePure64-10.1 (copy).iso"),
-	}
-	for _, path := range checklist {
-		if _, err = os.Stat(path); err != nil {
-			t.Fatalf("Get wrong cache directory, do not find file %s", path)
-		}
+func TestGoBackDir(t *testing.T) {
```
pallaud, line 136: Sorry I think I phrased my last comment incorrectly, I didn't mean take out the UI testing - I mean can you just test a simpler case where you call exec for a DirOption at the root level, go two levels down, choose the back option, and then confirm that the new DirOption that you're presenting is the correct one? or something like that

cmds/webboot/types.go
xsy1996: oh, so Ron suggest to use this code to replace all these "expected to implement ..." comment

cmds/webboot/webboot.go
xsy1996: You are right. filepath.Split is exactly equals to filepath.Dir +filepath.Base, and i just want the Dir

cmds/webboot/webboot.go
xsy1996: hmm I need this at the line 225 for check if the current dit of DirOption is the cache dir itself. if it is I should back to the main menu instead of go to it's father folder

cmds/webboot/webboot.go
xsy1996: no it just be called when the current path is equal to the path of cache dir.  since at that time , if I still go back to the father folder, I will be out side of the cache dir, which is not what I want. And at that time , I should back to the main menu right? Otherwise when the current fir is not the root of cache dir it will go back to it's father folder

xsy1996: May be I should record a video to show this process

cmds/webboot/webboot_test.go
xsy1996: Yes, I understand what you mean. but when I think it further, I find why I implementing the whole process of back option at test? Actually the process is, if a backOption is called, I return a DirOption with it's father's path right. and most of the options in this process have been tested in the previous test, like correctly call a Option in the menu is be tested by menu_test. and create a DirOption with certern path is be tested in the TestDirOption. So the rest logic that haven't be covered in the past test is only the goBack function. So I just test this function here is ok. Is that make sense?

xsy1996: And I think that makes the test more easily to read and understand.

cmds/webboot/webboot.go
rminnich: What is the right one to have? Which one is preferred?

cmds/webboot/webboot.go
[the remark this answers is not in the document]
rminnich: back is hard to remember. Is an empty string possible to make it go back? That's more common.

cmds/webboot/types.go
rminnich: yes, it's a variable declaration.
var _ = Entry(&ISO{})

cmds/webboot/types.go
rminnich: It's a very common idiom for ensuring that your type implements the interface.

cmds/webboot/webboot.go
rminnich: nevertheless the zero value of a string is ""
var cacheDir
is the same as 
var cacheDir=""

xsy1996: get it! thanks

cmds/webboot/webboot.go
xsy1996: I think the clean one is what the filepath want me to use in this situation, but for here I do not want it return "." when the dir is empty. Currently I still use clean, but I move the clean sentence inside to the if statement , so the if statement have checked that the dir will not be empty.

cmds/webboot/webboot.go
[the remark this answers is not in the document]
xsy1996: I think that's a good idea

cmds/webboot/webboot.go
pallaud: Actually I'm not sure that this variable is being used anymore, there's also a cacheDir in the main function which seems to be what you're using, right?

cmds/webboot/webboot.go
pallaud: Got it, I see what I misunderstood - could you explain why you break after calling getMainMenu? If you break then how is the entry that comes back from getMainMenu handled?

cmds/webboot/webboot_test.go
pallaud: I think it'd just be nice to have the UI test for this case, up to you :)

cmds/webboot/webboot.go
xsy1996: since getMainMenu(cacheDir) will returns the entry chooed in th main menu. and I user entry = getMainMenu(cacheDir) get this return. that is the same as what happened in other case: that is,  in every turn, I update my entry variable from the last turn's choose to the current tern's choose and  then begin the next turn.

cmds/webboot/webboot_test.go
xsy1996: emmm ok, I'll add the ui test too.

cmds/webboot/webboot.go
xsy1996: yeah I did not need it anymore... thank you Prachi!

cmds/webboot/webboot.go
```diff
@@ -119,39 +123,82 @@ func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
-		entries = append(entries, &DirOption{
-			label: "Use Cached ISO",
-			path:  *dir,
-		})
 	}
+	return "", fmt.Errorf("Do not find the cache directory: Expected a /Images under at the root of a block device(USB)")
+}
 
-	entry, err := menu.DisplayMenu("Webboot", "Choose an ISO:", entries, ui.PollEvents())
+func getMainMenu(cacheDir string) menu.Entry {
```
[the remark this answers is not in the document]
xsy1996: I have removed that global variable cacheDir

cmds/webboot/webboot.go
```diff
@@ -119,39 +123,82 @@ func (d *DirOption) exec(uiEvents <-chan ui.Event) (menu.Entry, error) {
...
+	entries = append(entries, &DownloadOption{})
+
+	entry, err := menu.DisplayMenu("Webboot", "Choose an option:", entries, ui.PollEvents())
 	if err != nil {
 		log.Fatal(err)
 	}
+	return entry
+}
+
+func goBackDir(currentPath string) menu.Entry {
```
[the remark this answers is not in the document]
xsy1996: Oh yeah. this function is just for test. But since I restored the test of backOption, it function is not useful anymore.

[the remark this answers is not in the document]
pallaud: Ok if you're still using it could we change the name to something like `getParentDir` so its a little clearer what its doing?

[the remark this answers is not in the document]
xsy1996: oh I just removed that. I think you are right it's useless.

PR #114, opened 2020-08-11
Removed config for compiling kernel v4.12.7

GBA-17:
Kernel v4.12.7 is a bit outdated, so we were hoping to switch to v5.6.14 as the stable kernel.

This commit removes the config file for compiling v4.12.7. We will make the changes for using v5.6.14 as the stable kernel in a separate PR.


GBA-17:
The commit should be signed now.

PR #115, opened 2020-08-11
Use kernel v5.6.14 as the stable version

GBA-17:
Update the instructions and config file for building kernel v5.6.14

Originally I was having trouble getting wifi to work on my local hardware. It turns out that I had to add the firmware for my wireless card to the `config-5.6.14` file.

Once I figured that out, I updated the config file and the instructions in the README to explain how I fixed the problem.


GBA-17:
The commit should be signed now.

PR #116, opened 2020-08-11
Add Travis CI

GBA-17:
Create Travis file and a shell script to run tests.

Currently we test the following:

- Build from root directory
- Create initramfs
- Unit tests for pkg/bootiso

Additional tests can be added by editing `travis.sh`. Note that if this PR is accepted, we still need to enable Travis from their website (otherwise the files are in the repo but never run).


travis.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
rminnich, line 2: if you put a
set -e
here, it will force an exit on any errors, if that helps. 

Nice to have this back.

GBA-17: Good idea, I just added `set -e`

GBA-17:
I think I made a mistake when I tried squashing my commits. Somehow some commits from the master branch got added in, so now the PR includes some of the changes that fixed dependency issues.

If you look at the [commit history](https://github.com/u-root/webboot/pull/116/commits) for the PR, the only work that should really be here starts at `Create .travis.yml file`

GBA-17:
I did a hard reset of the branch and manually copied over the files, so the PR should be more accurate now.

PR #117, opened 2020-08-12
Build  a scrollable menu

xsy1996:
page up : 'pageUp' key or 'leftArrow' key
page down : 'pageDown' key or 'rightArrow' key
move one line up: 'upArrow' key
move one line down: 'downArrow' key
first page: 'home' key
last page: 'end' key

### DIFF
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -11,6 +11,7 @@ import (
 )
 
 const menuWidth = 50
+const menuHeight = 12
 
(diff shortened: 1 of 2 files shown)


xsy1996:
te
cinnamon-20200812-16 (1)
st

xsy1996:
> so nice to see all the testing, but a scrollbar is great to have!

ok I'll think about how to add a scroll bar

PR #118, opened 2020-08-13
Cache downloaded iso

xsy1996:
If the cache dir is not find, downloaded the iso to /tmp， elsewise create a Downloaded dir under cache dir and download there.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -50,7 +50,7 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
 // DownloadOption's exec lets user input the name of the iso they want
 // if this iso is existed in the bookmark, use it's url
 // elsewise ask for a download link
-func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool) (menu.Entry, error) {
+func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool, cacheDir string) (menu.Entry, error) {
 	if network {
 		for {
 			ok, err := setUpNetwork(uiEvents)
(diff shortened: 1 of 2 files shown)


cmds/webboot/webboot.go
```diff
@@ -80,8 +80,19 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool) (menu.Entr
...
 	}
-
-	fpath := filepath.Join("/tmp", filename)
+	// If the cachedir is not find, downloaded the iso to /tmp, else create a Downloaded dir in the cache dir.
+	var fpath string
+	if cacheDir == "" {
+		fpath = filepath.Join("/tmp", filename)
+	} else {
+		downloadDir := filepath.Join(cacheDir, "Downloaded")
+		if _, err := os.Stat(downloadDir); os.IsNotExist(err) {
```
rminnich, line 89: just do an os.MkDirall here, it does all this stuff for you.

cmds/webboot/webboot.go
```diff
@@ -80,8 +80,19 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool) (menu.Entr
 			return nil, err
 		}
 	}
-
-	fpath := filepath.Join("/tmp", filename)
+	// If the cachedir is not find, downloaded the iso to /tmp, else create a Downloaded dir in the cache dir.
+	var fpath string
+	if cacheDir == "" {
+		fpath = filepath.Join("/tmp", filename)
```
rminnich, line 86: suggestion: instead of /tmp, use os.TempDir, that way your files won't conflict if you run this twice?

cmds/webboot/webboot.go
```diff
@@ -177,6 +188,10 @@ func main() {
 		verbose = log.Printf
 	}
 	cacheDir := *dir
+	if _, err := os.Stat(cacheDir); os.IsNotExist(err) {
+		log.Println("file does not exist")
```
rminnich, line 192: Where will this print come out? Will it mess up the UI? Does the UI have a way to present messages like this?

cmds/webboot/webboot.go
```diff
@@ -197,11 +212,11 @@ func main() {
 	for entry != nil {
 		switch entry.(type) {
 		case *DownloadOption:
-			if entry, err = entry.(*DownloadOption).exec(ui.PollEvents(), *network); err != nil {
+			if entry, err = entry.(*DownloadOption).exec(ui.PollEvents(), *network, cacheDir); err != nil {
 				log.Fatalf("Download option failed:%v", err)
 			}
 		case *ISO:
-			if err = entry.(*ISO).exec(ui.PollEvents(), *dryRun); err != nil {
+			if err = entry.(*ISO).exec(ui.PollEvents(), !(*dryRun)); err != nil {
```
rminnich, line 219: you can !*dryRun I thought? I doubt you need the ()

cmds/webboot/webboot.go
xsy1996: I think the message will be print it the original terminal window that we run the webboot program, and after the program is exit we can see this message.

xsy1996:
demo of use non exist cache dir.
cinnamon-20200813-3

PR #119, opened 2020-08-13
Enable mousewheel for scrolling

xsy1996:
Enable mouse wheel & Add a text scroll bar

I did not find a way to add a graphic scroll bar in termui, But I think this dynamic text could do the similar thing. Just allow user know where they are in this menu.


xsy1996:
the text scroll bar looks like this
cinnamon-20200813-2

PR #120, opened 2020-08-14
Add wifi setup functionality

GBA-17:
Add a menu system for setting up a wifi connection. Most of the back-end functionality for configuring the connection comes from [NiChrome's wifi package](https://github.com/u-root/NiChrome/tree/master/pkg/wifi).

We need to collect the following from the user in order to set up the network:

- Wireless network interface
- Wifi network name (ESSID)
- Login credentials for the wifi network

Most of the new logic is contained in the new `network.go` file, but we also modified the following:

- `types.go` in order to add menu entry classes for Interface and Network
- `webboot.go` to change the network setup function from the old `setUpNetwork()` to the new `setupNetwork()`
- `utils.go` to remove the old `setUpNetwork()` function

### DIFF
--- a/cmds/webboot/network.go
+++ b/cmds/webboot/network.go
@@ -1,0 +1,153 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	ui "github.com/gizak/termui/v3"
+	"github.com/u-root/NiChrome/pkg/wifi"
+	"github.com/u-root/webboot/pkg/menu"
+	"github.com/vishvananda/netlink"
+)
+
+func connected() bool {
+	client := http.Client{
+		Timeout: 10 * time.Second,
+	}
+
+	if _, err := client.Get("http://google.com"); err != nil {
+		return false
+	}
+	return true
+}
+
+func interfaceEntries() ([]menu.Entry, error) {
+	interfaces, err := netlink.LinkList()
+	if err != nil {
+		return nil, err
+	}
+
+	var ifEntries []menu.Entry
+	for _, iface := range interfaces {
+		ifEntries = append(ifEntries, &Interface{label: iface.Attrs().Name})
+	}
+	return ifEntries, nil
+}
+
+func interfaceIsWireless(ifname string) bool {
+	devPath := fmt.Sprintf("/sys/class/net/%s/wireless", ifname)
+	if _, err := os.Stat(devPath); err != nil {
+		return false
+	}
+	return true
+}
+
+func setupNetwork(uiEvents <-chan ui.Event) error {
+	iface, err := selectNetworkInterface(uiEvents)
+	if err != nil {
+		return err
+	}
+
+	return selectWirelessNetwork(uiEvents, iface)
+}
+
+func selectNetworkInterface(uiEvents <-chan ui.Event) (string, error) {
+	ifEntries, err := interfaceEntries()
+	if err != nil {
+		return "", err
+	}
+
+	for {
+		iface, err := menu.DisplayMenu("Network Interfaces", "Choose an option", ifEntries, uiEvents)
+		if err != nil {
+			return "", err
+		}
+
+		if !interfaceIsWireless(iface.Label()) {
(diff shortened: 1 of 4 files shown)


GBA-17:
There were some conflicts with the master branch, but they should be resolved now.

cmds/webboot/network.go
```diff
@@ -0,0 +1,151 @@
...
+	"os"
+	"time"
+
+	ui "github.com/gizak/termui/v3"
+	"github.com/u-root/NiChrome/pkg/wifi"
+	"github.com/u-root/webboot/pkg/menu"
+	"github.com/vishvananda/netlink"
+)
+
+func connected() bool {
```
xsy1996, line 15: Just in my opinion, I think you can have a new pkg and move functions that mostly focus on network things to there. But remain functions that mostly focus on UI there. Like he connected(), interfaceIsWireless() .etc could be moved to the new pkg

cmds/webboot/network.go
```diff
@@ -0,0 +1,151 @@
...
+
+func interfaceIsWireless(ifname string) bool {
+	devPath := fmt.Sprintf("/sys/class/net/%s/wireless", ifname)
+	if _, err := os.Stat(devPath); err != nil {
+		return false
+	}
+	return true
+}
+
+func setupNetwork(uiEvents <-chan ui.Event) (bool, error) {
```
xsy1996, line 47: I wondered would that be better if you do not call selectNetworkInterface inside the function, but send the name of interface as a parameter to this function? since the selectNetworkInterface is more focus on UI things, and the setupNetwork is actually more care about setting network. so after separating them this function could get rid of UI things. And maybe you can move this function to the new pkg.

cmds/webboot/network.go
```diff
@@ -0,0 +1,151 @@
...
+		if err != nil {
+			return err
+		}
+
+		network, ok := entry.(*Network)
+		if !ok {
+			return fmt.Errorf("Bad menu entry.")
+		}
+
+		var setupParams = []string{network.info.Essid}
```
xsy1996, line 108: I wondered would it be better if you separate this function to be two from this line. since all the logic above is to get a wifi choice from user, and all the logic below is to connect the wifi. And if you want to separate, the first function could return an network Entry instead of returning a iface string, so that you can add back option to the wifi menu.

cmds/webboot/network.go
```diff
@@ -0,0 +1,151 @@
...
+
+		if !interfaceIsWireless(iface.Label()) {
+			menu.DisplayResult([]string{"Only wireless network interfaces are supported."}, uiEvents)
+		} else {
+			return iface.Label(), nil
+		}
+	}
+}
+
+func selectWirelessNetwork(uiEvents <-chan ui.Event, iface string) error {
```
xsy1996, line 81: I am not sure, but did you provide a way for user to input the password of network? it that only support non password wifi now?

cmds/webboot/network.go
```diff
@@ -0,0 +1,151 @@
...
+	return true
+}
+
+func setupNetwork(uiEvents <-chan ui.Event) (bool, error) {
+	iface, err := selectNetworkInterface(uiEvents)
+	if err != nil {
+		return false, err
+	}
+
+	err = selectWirelessNetwork(uiEvents, iface)
```
pallaud, line 53: Nit: you could also directly do `return selectWirelessNetwork(uiEvents, iface)` since you're not modifying the error you get back

GBA-17: That makes sense, I updated `selectWirelessNetwork()` with that change.

cmds/webboot/network.go
GBA-17: I think that's a good idea, I separated that portion into a new function `connectWirelessNetwork()`.

cmds/webboot/network.go
GBA-17: I'm not sure it's worth making a separate package yet. `connected()` and `interfaceIsWireless()` total to about 10 lines, and then the rest of the code involves the menu flow from the main program.

GBA-17:
I ended up resetting the branch and copying over the files, so everything should be in a single commit now. Sorry again for all the new commits.

PR #121, opened 2020-08-14
Add back option to Download.

xsy1996:
Enter an empty string to go back.


xsy1996:
Demo
cinnamon-20200814-5

PR #124, opened 2020-08-19
Add page turning feature in DisplayResult

xsy1996:
Press any key to continue, press <Esc> to exit.


xsy1996:
cinnamon-20200819-1

PR #126, opened 2020-08-20
Fix UI issues

xsy1996:
1.Add "press ctrl-d to exit" message in main menu.
2.Persist the warning message.
3.Allow regexp kind of input in download. If the input is match one and only one
bookmark, assume this is what the user want. If the input match more
than one bookmark, print all matched bookmark in warning.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	ui "github.com/gizak/termui/v3"
 	"github.com/u-root/u-root/pkg/mount"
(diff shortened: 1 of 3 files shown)


cmds/webboot/webboot.go
```diff
@@ -178,7 +193,7 @@ func getMainMenu(cacheDir string) menu.Entry {
 	}
 	entries = append(entries, &DownloadOption{})
 
-	entry, err := menu.DisplayMenu("Webboot", "Choose an option:", entries, ui.PollEvents())
+	entry, err := menu.DisplayMenu("Webboot", "Choose an option:(press Ctrl-d to exit)", entries, ui.PollEvents())
```
rminnich, line 196: very very nice :-)

PR #135, opened 2020-08-25
Avoid closing UI and showing output from the command line.

GBA-17:
One issue with the UI was that when menu screens changed, output from the command line sometimes showed up on screen. This might be confusing to a user, who might think that the program crashed.

Avoiding this required 2 changes:

1. Only initialize and close termui once. If we need to change screens, call termui's `Clear()` function.
2. Supress output from `log` and `fmt`. Even if we avoid closing termui, text from the output can still overlap with the menu.

With this change applied, menu screens will hang when the user makes a selection that triggers a long-running operation (ex. wifi scanning, downloading). The next PR will include a loading animation that runs during these operations.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -233,9 +234,9 @@ func main() {
 	}
 	defer menu.Close()
 
-	// Suppress output, else it might overlap with the menu
-	log.SetOutput(ioutil.Discard)
-	os.Stdout = nil
+	// Buffer the log output, else it might overlap with the menu
+	var logBuffer bytes.Buffer
+	log.SetOutput(&logBuffer)
 
 	entry := getMainMenu(cacheDir)
 
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -227,6 +227,16 @@ func main() {
 			cacheDir = cachePath
 		}
 	}
+
+	if err := menu.Init(); err != nil {
+		log.Fatalf(err.Error())
+	}
+	defer menu.Close()
+
+	// Suppress output, else it might overlap with the menu
+	log.SetOutput(ioutil.Discard)
+	os.Stdout = nil
+
 	entry := getMainMenu(cacheDir)
 
 	var err error
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -37,6 +37,18 @@ func max(a, b int) int {
 	return b
 }
 
+func Init() error {
+	return ui.Init()
+}
+
+func Clear() {
+	ui.Clear()
+}
+
+func Close() {
+	ui.Close()
+}
+
 // AlwaysValid is a special isValid function that check nothing
 func AlwaysValid(input string) (string, string, bool) {
 	return input, "", true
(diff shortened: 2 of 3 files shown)


cmds/webboot/webboot.go
```diff
@@ -227,6 +227,16 @@ func main() {
 			cacheDir = cachePath
 		}
 	}
+
+	if err := menu.Init(); err != nil {
+		log.Fatalf(err.Error())
+	}
+	defer menu.Close()
+
+	// Suppress output, else it might overlap with the menu
```
pallaud, line 236: I wonder if there's a way we could just defer the output to after the menu exists rather than discarding it in case the menu closes with an error and we want to see the log output?

pkg/menu/menu.go
```diff
@@ -37,6 +37,18 @@ func max(a, b int) int {
 	return b
 }
 
+func Init() error {
+	return ui.Init()
+}
+
+func Clear() {
```
pallaud, line 44: I don't see this being used anywhere?

cmds/webboot/webboot.go
GBA-17: That's a good point. Maybe I could write it to a file?

pkg/menu/menu.go
GBA-17: I'll take that out. I think the menus already clear themselves with `ui.Clear()`.

cmds/webboot/webboot.go
rminnich: Rule 1: no discarding output
Rule 2: no writing it to random files. You can put it in memory.

you can do this for logging. 

r, w := io.Pipe
log.SetOutput(w)
the log output will appear in the pipe. you can get it later by reading from r.

also this os.Stdout? Why do this? never do this IOW and don't plan on using files ...
where is the code that writes to os.Stdout? A command you spawn? Let's talk some more. 
Let's do one thing at a time however. Can you fix the log bits first?

GBA-17: I'll make that change for the log output. Should we do anything with `r` at the end?

Also, the only thing that gets written to `os.Stdout` is the download progress.

GBA-17: I made the change locally, but the program hangs at the function call that generates the logs. I think the problem might have to do with the pipe being synchronous.

Would it be ok to make a Writer class with an internal buffer?

rminnich: use a bytes.Buffer instead of the io.Pipe. Sorry, I have you bad advice.

GBA-17: No problem, it should be updated now.

GBA-17:
I think I'm still having a problem with the log output. Redirecting to a byte buffer prevented any output from `syslinux.ParseLocalConfig()`, but I'm still getting output to my console from `wifi.Connect()` when it calls `dhclient.SendRequests()`. 

It looks like both of them call `log.Printf()`, so I'm sure why logging only stops for one of them.

GBA-17:
It turns out the problem with `wifi.Connect()` was that it runs a few commands with `exec()`. Originally it would redirect the `exec()` output to Stdout for debugging purposes, so I took those lines out.

cmds/webboot/webboot.go
```diff
@@ -227,6 +228,16 @@ func main() {
 			cacheDir = cachePath
 		}
 	}
+
+	if err := menu.Init(); err != nil {
```
rminnich, line 232: could menu.Init ever have error messages you want to buffer too?

pkg/menu/menu.go
```diff
@@ -119,27 +127,18 @@ func NewInputWindow(introwords string, isValid validCheck, uiEvents <-chan ui.Ev
 
 // NewCustomInputWindow creates a new ui window and displays an input box.
 func NewCustomInputWindow(introwords string, wid int, ht int, isValid validCheck, uiEvents <-chan ui.Event) (string, error) {
```
rminnich, line 129: I'm wondering if the New and the Display should not be two things. 
New would give the window and the user would then call input processing. 
THink about how exec.Command and exec.Run or exec.Start do this.

cmds/webboot/webboot.go
GBA-17: I don't think so in this case. `menu.Init()` really just wraps `termui.Init()`, and [the only return value](https://github.com/gizak/termui/blob/master/v3/backend.go#L13) is an error. 

I think if the UI doesn't init, the program should terminate.

pkg/menu/menu.go
GBA-17: I think I like the way April set it up, just because it feels like how other languages collect user input. It feels like doing something like `val = input("Enter your value: ")`, and then just waiting until the user inputs something before continuing.

Except I also like how the user can customize the exec options between `exec.Command()` and `exec.Run()`. Maybe the user wants to set a custom validation function on the input box.

Maybe the compromise is to have some widgets that are entirely customizable by the caller, and others that are just like a `val = input()` function call. CustomInputWindow would be like the latter, but something like the Progress widget in [the new PR](https://github.com/u-root/webboot/pull/136) gives the caller more control.

PR #136, opened 2020-08-25
Create progress screen.

GBA-17:
Created a progress screen to display during long-running operations (ex. wifi scanning, downloads). The screen is animated so that users don't feel like the program has stopped running.

There are 2 modes of animation:

1. Blinking dots provided by the asynchronous `animate()` method
2. Manual updates from the caller. This would be useful for showing the download progress, which updates multiple times per second.

The menu is used in the following way:

1. Create a Progress object `p` by calling `NewProgress()`. The caller must specify the initial text and whether they would like the default animation. 
2. Run the long-running operation. If the caller did not use the default animation, update the text manually by calling `p.Update()`
3. When the caller is done with `p`, call the method `p.Close()`

### DIFF
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"strconv"
 	"strings"
+	"time"
 
 	ui "github.com/gizak/termui/v3"
 	"github.com/gizak/termui/v3/widgets"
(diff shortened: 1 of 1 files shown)


GBA-17:
Peek 2020-08-25 11-20

Sample of the default animation.

GBA-17:
I noticed that most long-running operations output to the log, so I was planning to include the actual usage once the logging PR is finalized.

Would it be better if I included the files where `Progress` is used?

PR #137, opened 2020-08-26
Buffer output from the wifi Connect() function.

GBA-17:
`wifi.Connect()` executes 2 shell commands and redirects the output to Stdout/Stderr. This output tends to overlap with the UI, so we would like to avoid showing it.

In this PR, we store the output from the commands in a `bytes.Buffer`. This allows us to keep it for later access, while keeping it from messing up the UI.

### DIFF
--- a/pkg/wifi/iwl.go
+++ b/pkg/wifi/iwl.go
@@ -5,10 +5,10 @@
 package wifi
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"os/exec"
 	"regexp"
 	"strings"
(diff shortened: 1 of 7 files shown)


pkg/wifi/iwl.go
```diff
@@ -155,14 +155,16 @@ func (w *IWLWorker) Connect(a ...string) error {
 	// it's been almost instantaneous. But, further, it needs to keep running.
 	go func() {
 		cmd := exec.CommandContext(ctx, "wpa_supplicant", "-i"+w.Interface, "-c/tmp/wifi.conf")
-		cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr //For an easier time debugging
+		var outBuffer, errBuffer bytes.Buffer
```
rminnich, line 158: This was not my suggestion. My suggestion was that your funnction take an io.Writer as a parameter, e.g.
go func(w io.Writer)
and then 
cmd.Stdout, cmd.Stderr = w, w

and the io.Writer can be anything: an os.Pipe, an io.Pipe, a bytes.Buffer, does not matter

pkg/wifi/iwl.go
```diff
@@ -155,14 +155,16 @@ func (w *IWLWorker) Connect(a ...string) error {
...
+		cmd.Stdout, cmd.Stderr = &outBuffer, &errBuffer //For an easier time debugging
 		cmd.Run()
 	}()
 
 	// dhclient might never return on incorrect passwords or identity
 	go func() {
 		cmd := exec.CommandContext(ctx, "dhclient", "-ipv4=true", "-ipv6=false", "-v", w.Interface)
-		cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr //For an easier time debugging
+		var outBuffer, errBuffer bytes.Buffer
```
rminnich, line 166: same here. Pass in an io.Writer (or two, if you want to keep cmd.Stdout and cmd.Stderr separate) -- up to you

pkg/wifi/iwl.go
GBA-17: That makes sense. Would you recommend using a separate io.Writer for each cmd, or can they share the same one?

GBA-17:
Got it. I changed the API for the `wifi` class, so now any function that runs `exec()` on something writes to a Stdout and Stderr Writer.

In webboot's `network.go` file, I declared a global Stdout and Stderr buffer to record output from the network setup process.

PR #138, opened 2020-08-27
Write log output to a bytes.Buffer

GBA-17:
The log output from `syslinux.ParseLocalConfig()` would sometimes overlap with the UI. Instead of writing logs to the screen, this PR creates a `bytes.Buffer` and writes the log output to the buffer.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


cmds/webboot/webboot.go
```diff
@@ -229,10 +230,14 @@ func main() {
 	}
 	entry := getMainMenu(cacheDir)
 
-	var err error
+	// Buffer the log output, else it might overlap with the menu
+	var logBuffer bytes.Buffer
```
rminnich, line 234: do you have a way to display this output anywhere? is this a bitbucket for errors?

GBA-17: Right now it's just collecting things that get written to the log. I think the only function that writes to the log is `syslinux.ParseLocalConfig()`, and all it displayed was a formatted list of the boot configurations it found.

GBA-17: In the future, maybe we could pass the buffer to the `Progress` widget and have it display the output live. For now though, I was just buffering the output and storing it for a later use case.

PR #146, opened 2020-09-02
Add function to infer type of cached ISO

GBA-17:
`supportedDistros` contains the boot parameters for the distributions we officially support. This PR adds the function `inferIsoType()` that attempts to use a cached ISO's name to match it to an entry in `supportedDistros`.

This involved adding a regex to the `Distro` struct that contains each distribution's ISO naming convention.

### DIFF
--- a/cmds/webboot/types.go
+++ b/cmds/webboot/types.go
@@ -9,6 +9,7 @@ import (
 
 type Distro struct {
 	url          string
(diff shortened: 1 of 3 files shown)


rminnich (approved):
not sure I see the big picture yet but I'll take your word for it :-)

PR #154, opened 2020-09-03
Show error messages when errors happening

rminnich:
It's impossibly frustrating to have them go into a black hole.
This will do until we can do better.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -220,8 +220,8 @@ func main() {
 	entry := getMainMenu(cacheDir)
 
 	// Buffer the log output, else it might overlap with the menu
-	var logBuffer bytes.Buffer
-	log.SetOutput(&logBuffer)
+	var l bytes.Buffer
+	log.SetOutput(&l)
 
 	// check the chosen entry of each level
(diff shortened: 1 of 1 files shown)


GBA-17 (approved):
This looks good to me.

I think it might also be worth printing out `wifiStdout` and `wifiStderr` if the Download option fails.

rminnich:
it's actually not good. It should print that message and go back to the dialog, not blow out of the program the way it does. The error message handling has some real problems.

xsy1996:
I am not really understand.. why we do not use menu.DisplayResult to display the error message?

xsy1996:
I think you may use menu.DisplayResult to dispaly the error, and then go back to the upper menu just act the same as we hit the back opion.

for example when download option is error:

case *DownloadOption:
			if entry, err = entry.(*DownloadOption).exec(ui.PollEvents(), *network, cacheDir); err != nil {
				menu.DisplayResult([]string{fmt.Sprintf("Download option failed:%v (%s)", err, l.String())}, ui.PollEvents())
				entry = getMainMenu(cacheDir)
			}
			if _, ok := entry.(*BackOption); ok {
				entry = getMainMenu(cacheDir)
			}


dose that work?

PR #156, opened 2020-09-08
Added function to boot cached ISO

GBA-17:
Adds the function `BootCachedISO()` which kexec's into the new distribution. Unlike the earlier `BootFromPmem()` function, no pmem devices are accessed. Instead, we expect the caller to send over kernel parameters related to the location of the ISO file.

For example, if we have an Ubuntu iso at `/Images/ubuntu.iso`, we append `iso-scan/filename=/Images/ubuntu.iso` to the kernel parameters in the kexec call.

With this patch applied, we can now boot a local Ubuntu ISO.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"sort"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


rminnich (approved):

rminnich:
IT WORKS!!!!

PR #160, opened 2020-09-10
Allow numbered menu to return BackOption if <Esc> is pressed

GBA-17:
Backtracking from a numbered menu was implemented previously in the `DirOption` exec function. The implementation required manually adding a `BackOption` to the list of directory entries, and having logic to handle the selection of a `BackOption`.

This idea was extended to all numbered menus with the following changes:

- Show text at the bottom of each menu to indicate that \<Esc> returns to an earlier menu
- Move the `BackOption` struct to the menu package
- If the user types \<Esc> in a numbered menu, return a `BackOption`
- Ensure that any user-created numbered menu can properly handle a `BackOption`

### DIFF
--- a/cmds/webboot/network.go
+++ b/cmds/webboot/network.go
@@ -52,33 +52,35 @@ func interfaceIsWireless(ifname string) bool {
 	return true
 }
 
-func setupNetwork(uiEvents <-chan ui.Event) error {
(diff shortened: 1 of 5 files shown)


GBA-17:
Screenshot from 2020-09-10 12-46-11

Controls are now listed at the bottom of the menu.

rminnich (approved):

PR #161, opened 2020-09-11
Handle <Esc> press from text inputs

GBA-17:
The `PromptTextInput()` function originally returned \<Esc> when the user pressed the escape key, and io.EOF when they pressed \<Ctrl+d>. Instead of checking for \<Esc>, we can return a custom error type `BackRequest` and add logic to each caller to handle the new error type.

We currently only have text input boxes in the section that collects wifi credentials, so the changes are mostly limited to `network.go`

### DIFF
--- a/cmds/webboot/network.go
+++ b/cmds/webboot/network.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"time"
@@ -108,8 +109,13 @@ func selectWirelessNetwork(uiEvents <-chan ui.Event, iface string) (bool, error)
 			return false, fmt.Errorf("Bad menu entry.")
 		}
 
-		if err := connectWirelessNetwork(uiEvents, worker, network.info); err != nil {
+		completed, err := connectWirelessNetwork(uiEvents, worker, network.info)
+		if err == io.EOF { // user typed <Ctrl+d> to exit
+			return false, err
+		} else if err != nil { // connection error
 			menu.DisplayResult([]string{err.Error()}, uiEvents)
+			continue
+		} else if !completed { // user typed <Esc> to go back
 			continue
 		}
 
@@ -117,16 +123,18 @@ func selectWirelessNetwork(uiEvents <-chan ui.Event, iface string) (bool, error)
 	}
 }
 
-func connectWirelessNetwork(uiEvents <-chan ui.Event, worker wifi.WiFi, network wifi.Option) error {
+func connectWirelessNetwork(uiEvents <-chan ui.Event, worker wifi.WiFi, network wifi.Option) (bool, error) {
 	var setupParams = []string{network.Essid}
 	authSuite := network.AuthSuite
 
 	if authSuite == wifi.NotSupportedProto {
-		return fmt.Errorf("Security protocol is not supported.")
+		return false, fmt.Errorf("Security protocol is not supported.")
 	} else if authSuite == wifi.WpaPsk || authSuite == wifi.WpaEap {
 		credentials, err := enterCredentials(uiEvents, authSuite)
 		if err != nil {
-			return err
+			return false, err
+		} else if credentials == nil {
+			return false, nil
 		}
 		setupParams = append(setupParams, credentials...)
 	}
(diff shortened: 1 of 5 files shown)


cmds/webboot/network.go
```diff
@@ -157,6 +167,8 @@ func enterCredentials(uiEvents <-chan ui.Event, authSuite wifi.SecProto) ([]stri
 	identity, err := menu.PromptTextInput("Enter identity:", menu.AlwaysValid, uiEvents)
 	if err != nil {
 		return nil, err
+	} else if identity == "<Esc>" {
```
pallaud, line 170: I wonder if it'd make sense to create a new error type for back option and then switch on the error instead of having to check for esc? I think then you'd also be able to avoid adding the extra bool to the return type. Not sure, what do you think?

GBA-17: Yeah I really like that idea.

Actually, that would also help make the return type for `PromptTextInput()` more consistent with the one for `PromptMenuEntry()`.

I'll make that change here, and I might have to add another PR to make some changes to `PromptMenuEntry()`.

xsy1996: Yeah I should have designed a special error for <Escape> at the very beginning  instead of using the return content to recognize the <Escape> key. Just for reminding, not only the PromptTextInput() here but also the parsingMenuOption() is using the return content to recognize the <Escape> key and you may want to change all of them. Thank you for fix that!!

pkg/menu/menu.go
```diff
@@ -162,7 +162,7 @@ func DisplayResult(message []string, uiEvents <-chan ui.Event) (string, error) {
 	p.SetRect(0, 0, wid+2, resultHeight+3)
 	p.TextStyle.Fg = ui.ColorWhite
 
-	hint := "(Press any key to continue, press <Esc> to exit.)"
+	hint := "(Press any key to continue.)"
```
xsy1996, line 165: I wondered why remove "press <Esc> to exit" here?  <Esc> key in the DisplayResult() is doing different thing from in other two (input box and menu). In DisplayResult() <Esc> key is design for , if the alert message is so long that user have no patient to read that, they may want to use <Esc> to skip this message and immediately return to former window. But press any other key will just move to the next page of the message.

By the way, this is what I planed to do but have no time to: can you add scrollable feature to the message box? it will be very similar to the menu's input. After all pressing a key to turn the page may be not as convinient as just scrolling the wheel.

GBA-17: I don't think any of the callers to `DisplayResult()` are checking the return value right now, so I took it out for now. I'm planning to use Prachi's suggestion to return a custom error if the user presses \<Esc>, so I can update `DisplayResult()` with that change also.

Also, good idea with the scrollable message. I'll try to get that change done this week.

xsy1996: Sounds great!

cmds/webboot/types.go
```diff
@@ -132,3 +132,13 @@ func (n *Network) Label() string {
 	}
 	return "Invalid wifi network."
 }
+
+// BackOption is a special menu Entry that says
+// the user wants to go back to the previous menu
+type BackOption struct{}
```
xsy1996, line 138: Does it need backoption anymore? How about directly return the menu.BackRequest error? And the main function could recognize that user want to go back by this special error.

GBA-17: Actually, in the next PR I hoping to move `BackOption` back to `types.go`. I had moved it to the menu package when I was changing how `PromptMenuEntry()` would handle an \<Esc> press, but now I think we're passing back an error type when the user presses \<Esc>.

pallaud (approved):

cmds/webboot/types.go
xsy1996: I see, but my think is how about remove all backoption struct, both in menu and in types. At the beginning ,we define backoption is because we want there is a entry in menu for user to go back and we want it's format is close to other entry.  But for now, since we plan to use <Esc> to go back and the menu do not need to display the back entry any more,  in my mind the backoption's usage is just letting the main function know that user want to go back, right? Then, why not use the menu.BackRequest error to do the same thing so that simplify the code?

GBA-17: Yeah, actually that's a good idea. I was thinking of adding an error handler in the main loop at `webboot.go` where we can check the error types. That way we can handle back options, and also exit with 0 if the user pressed \<Ctrl+d>

xsy1996: Yeah good idea!

PR #162, opened 2020-09-11
Show percent downloaded in download progress

GBA-17:
The size of an incoming file is provided by `Response.ContentLength`. We can use this information to display the percentage of file downloaded. This involved the following changes:

- Update `linkOpen()` to return the expected file size. The function now supports only http and https.
- Update the `WriteCounter` object to include the fields `received` and `expected`
- Initialize `WriteCounter` with the expected file size
- Update the Sprintf expression to print the percent of file downloaded

### DIFF
--- a/cmds/webboot/utils.go
+++ b/cmds/webboot/utils.go
@@ -12,21 +12,21 @@ import (
 	"github.com/u-root/webboot/pkg/menu"
 )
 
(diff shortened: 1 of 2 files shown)


GBA-17:
Screenshot from 2020-09-11 10-31-29

Percent downloaded is now displayed, along with amount downloaded.

pallaud (approved):
Cool! :)

PR #167, opened 2020-09-16
Add error handler for the main program loop

GBA-17:
Most errors are currently handled by passing them to the main program loop, printing the error message and logs, and then exiting the program. Instead of exiting, this PR adds the function `handleError()` which prints the error message and logs to `menu.DisplayResult()`. When the user presses a key to exit the display page, we return to the main menu.

The error handler also checks for the `menu.ExitRequest` type, which indicates that the user pressed <Ctrl+d> to exit the program. In this case, we call `menu.Close()` and then exit the program with status 0.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -20,12 +20,13 @@ import (
 )
 
 var (
-	v        = flag.Bool("verbose", false, "Verbose output")
-	verbose  = func(string, ...interface{}) {}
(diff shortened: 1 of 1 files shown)


GBA-17:
Peek 2020-09-16 08-53

Sample error message and return to the main menu.

GBA-17:
Peek 2020-09-16 08-58

Sample of error screen and return to main menu,

pallaud (approved):

pallaud (approved):

PR #168, opened 2020-09-16
Scroll text input if value is longer than 80 characters

GBA-17:
Previously, if the value in a text input exceeded the size of the input box (80 characters), the overflow would not be displayed on screen.

This PR scrolls the text input box so that the last 80 characters of input are always displayed. It also adds a new test to ensure that long values are processed correctly.

### DIFF
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -87,37 +87,52 @@ func processInput(introwords string, location int, wid int, ht int, isValid vali
 	ui.Render(input)
 	ui.Render(warning)
 
-	// keep tracking all input from user
+	// The input box is wid characters wide
+	//   - 2 chars are reserved for the left and right borders
+	//   - 1 char is left empty at the end of input to visually
+	//        signify that the text box is still accepting input
+	// The user might want to input a string longer than wid-3
+	// characters, so we store the full typed input in fullText
+	// and display a substring of the full text to the user
+	var fullText string
+
 	for {
 		k := readKey(uiEvents)
 		switch k {
 		case "<C-d>":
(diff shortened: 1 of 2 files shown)


GBA-17:
Peek 2020-09-16 09-57

New text input.

GBA-17:
Peek 2020-09-16 10-01

Previous behavior.

GBA-17:
Peek 2020-09-16 10-03

Previous behavior.

pkg/menu/menu.go
```diff
@@ -88,36 +88,45 @@ func processInput(introwords string, location int, wid int, ht int, isValid vali
...
+			fullText = ""
 			input.Text = ""
 			warning.Text = warningString
 			ui.Render(input)
 			ui.Render(warning)
 		case "<Backspace>":
 			if len(input.Text) > 0 {
-				input.Text = input.Text[:len(input.Text)-1]
+				fullText = fullText[:len(fullText)-1]
+				start := max(0, len(fullText)-wid+3)
```
pallaud, line 111: Could you add a comment for future contributors explaining the numbers here? Like what is `wid` and why we add 3?

pkg/menu/menu.go
```diff
@@ -87,37 +87,52 @@ func processInput(introwords string, location int, wid int, ht int, isValid vali
 	ui.Render(input)
 	ui.Render(warning)
 
-	// keep tracking all input from user
+	// The input box is wid characters wide
```
GBA-17, line 90: Would this work for the comment?

pallaud (approved):

PR #169, opened 2020-09-16
Re-add custom URL option to the Download menu

GBA-17:
When we switched the download menu from a text input to a numbered menu, we removed the ability to type a custom URL. This PR re-adds that ability by doing the following:

- Add a new Distro option called "Custom Distro" to the bookmark menu
- If the user selects the custom option, prompt them for the URL

A new test was also added that selects the "Custom Distro" option and manually types out the URL for Tinycore.

### DIFF
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -119,6 +119,8 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool, cacheDir s
 	}
 
 	entries := supportedDistroEntries()
+	customLabel := "Other Distro"
+	entries = append(entries, &Config{customLabel})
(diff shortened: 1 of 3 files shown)


cmds/webboot/webboot.go
```diff
@@ -118,6 +118,7 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool, cacheDir s
 	}
 
 	entries := supportedDistroEntries()
+	entries = append(entries, &Config{"Other Distro"})
```
pallaud, line 121: Make "Other Distro" a constant?

GBA-17: Yeah that makes sense. It should be fixed in the new push.

pallaud (approved):

PR #171, opened 2020-09-18
Download ISO to temp file before writing to USB stick

GBA-17:
Previously, the `download()` function would create a file on the USB stick and write to the file during the download. This caused problems when the download was interrupted, since we would end up with a corrupted ISO file in the cache.

Instead of writing the download directly to the cache, this PR first writes the download to a temp file. The temp file is always removed at the end of the function.

If the download succeeds, the temp file is copied to the USB cache. Otherwise, if the download fails or is interrupted by the user, the temp file is just removed and the USB cache is not modified.

### DIFF
--- a/cmds/webboot/utils.go
+++ b/cmds/webboot/utils.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"regexp"
@@ -35,7 +36,8 @@ func (wc *WriteCounter) Close() {
 	wc.progress.Close()
 }
 
-// download will download a file from URL and save it as fPath
+// download() will download a file from URL and save it to a temp file
+// If the download succeeds, the temp file will be copied to fPath
 func download(URL, fPath string, uiEvents <-chan ui.Event) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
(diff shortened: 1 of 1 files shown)


rminnich (dismissed):
will the cancel work if it happens on the copy to the usb stick? Should there be some information printed when that starts?

GBA-17:
That's a good point, maybe we can run a second progress screen when the copy happens?

GBA-17:
I just added a progress screen for the copy step, but it is ok if the user can't cancel during that?

cmds/webboot/utils.go
```diff
@@ -51,17 +53,35 @@ func download(URL, fPath string, uiEvents <-chan ui.Event) error {
...
+		return err
+	}
+
+	cacheFile, err := os.Create(fPath)
+	if err != nil {
+		return err
+	}
+	defer cacheFile.Close()
+
+	if _, err := io.Copy(cacheFile, tempFile); err != nil {
```
rminnich, line 84: If this fails will it leave a partial file on the stick? If it fails should we remove the file?

cmds/webboot/utils.go
```diff
@@ -51,17 +53,35 @@ func download(URL, fPath string, uiEvents <-chan ui.Event) error {
...
+
+	counter.Close()
+	copyProgress := menu.NewProgress("Download complete. Writing ISO to cache", true)
+	defer copyProgress.Close()
+
+	if _, err = tempFile.Seek(0, 0); err != nil {
+		return err
+	}
+
+	cacheFile, err := os.Create(fPath)
```
rminnich, line 78: Should we check to see if the file is there? if it is there, what should we do?

GBA-17: Yeah, if the requested file name matches a file we already have, we can probably avoid downloading it.

Maybe we can check if the file already exists at the beginning of the function?

cmds/webboot/utils.go
GBA-17: Right now it would leave a partial file on disk. Maybe I can add a line to the error case to remove the file?

rminnich: yes sounds good

cmds/webboot/utils.go
rminnich: what if it is corrupt? If the system was reboot in the middle of copying to the usb, half a file could be there. Should we give them the option of keeping it or not?

rminnich: or could we do an md5sum and if it fails, remove it?

GBA-17: I think the md5sum is a good idea.

Should we remove the existing ISO by default, or maybe show a confirmation screen?

rminnich: I would confirm each step ....
let's see how that looks.

cmds/webboot/utils.go
GBA-17: I was thinking of adding `os.Remove(cacheFile.Name())` inside the error case, but would that cause any problems with the deferred `cacheFile.Close()`?

rminnich: no. In fact let's talk about why tomorrow, the reason is interesting.

GBA-17: Sounds good, I added the `os.Remove()` line.

cmds/webboot/utils.go
GBA-17: I think we can do the file exists checks outside of the download function. I'll put up a PR for that in a bit.

GBA-17:
I think this PR is mostly ready to go.

The only thing is that we don't check whether the file already exists, but I think we can do that outside of the download function.

pallaud (dismissed):
Approving this, and noting that the checks for whether a file exists will be added in a follow up PR.

pallaud (approved):

PR #172, opened 2020-09-18
Remove BackOption type, superseded by menu.BackRequest

GBA-17:
Remove the `BackOption` type, since it was superseded by the `menu.BackRequest` type.

This will make error handling simpler. In the past, we used to have error handlers to check for `menu.BackRequest` and return a `BackOption` if that error type was found. That looked something like this:

```go
entry, err := PromptTextEntry(...)
if err != nil {
    switch err {
    case menu.BackRequest:
        return &BackOption, nil
    default:
        return nil, err
}
```

By restructuring the program to recognize `menu.BackRequest`, we can achieve the same effect with the following:

```go
entry, err := PromptTextEntry(...)
if err != nil {
    return err
}
```

### DIFF
--- a/cmds/webboot/types.go
+++ b/cmds/webboot/types.go
@@ -132,13 +132,3 @@ func (n *Network) Label() string {
 	}
 	return "Invalid wifi network."
 }
-
-// BackOption is a special menu Entry that says
-// the user wants to go back to the previous menu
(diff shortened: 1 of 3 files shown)


rminnich (approved):
wow, this is nice.

PR #175, opened 2020-09-21
Added checksum validation function to the bootiso package

GBA-17:
Added a function to validate the checksum, given an ISO and its checksum file.

The checksum file provided with an ISO will look something like this:

```
2f6ae466ec9b7c6255e997b82f162ae88bfe640a8df16d3e2f495b6281120af9 *linuxmint-20-cinnamon-64bit.iso
42fd764b3a3544a36d820f4164bb64aa5a6d982073e6d1afdea4853d3858fc98 *linuxmint-20-mate-64bit.iso
761fb276da9746a068f4c8aa42e8d4981f352db92babe0ef8a08713eeb38246f *linuxmint-20-xfce-64bit.iso
```

We can parse out the checksum by finding the line that contains the ISO's filename. After calculating the ISO checksum ourselves, we can compare the calculated and parsed checksums.

A few tests are included to verify that the checksum function works as expected.

### DIFF
--- a/pkg/bootiso/bootiso.go
+++ b/pkg/bootiso/bootiso.go
@@ -1,11 +1,17 @@
 package bootiso
 
 import (
+	"bufio"
 	"context"
+	"crypto/md5"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"hash"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


orangecms (approved):
nice :)

PR #176, opened 2020-09-23
Add confirmation screen to the menu package

GBA-17:
We currently have menu screens for text entry, numbered options, and displaying error text. This PR adds a new screen that displays a message to the user, and asks them to reply yes or no. The function returns a boolean representing the result.

We can call the new menu function in the following way:

```go
confirm, err := menu.PromptConfirmation("Checksum could not be verified. Would you like to continue?", uiEvents)
```

This would result in the following menu, which would return `true` for 0 and `false` for 1.

Screenshot from 2020-09-23 12-29-08

### DIFF
--- a/pkg/menu/menu.go
+++ b/pkg/menu/menu.go
@@ -38,6 +38,16 @@ func max(a, b int) int {
 	return b
 }
 
+func countNewlines(str string) int {
+	count := 0
+	for _, s := range str {
+		if s == '\n' {
+			count++
+		}
+	}
+	return count
+}
+
 func Init() error {
 	return ui.Init()
 }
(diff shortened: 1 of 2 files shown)


pkg/menu/menu.go
```diff
@@ -38,6 +38,16 @@ func max(a, b int) int {
...
 
+func countNewlines(str string) int {
+	count := 0
+	for i := 0; i < len(str); i++ {
+		if str[i:i+1] == "\n" {
+			count++
+		}
+	}
+	return count
+}
```
rminnich, line 49: another way to count newlines
len(strings.Split(b[:i], "\n"))-1

https://play.golang.org/p/nuX1p_BGX7a

but in any event  use an iterator
for _, s := range str {
if s == '\n' {
count++
}
}

pkg/menu/menu.go
```diff
@@ -372,6 +382,64 @@ func PromptMenuEntry(menuTitle string, introwords string, entries []Entry, uiEve
...
+			}
+		case "0", "1":
+			entry.Text = key
+			ui.Render(entry)
+		case "<Backspace>":
+			entry.Text = ""
+			ui.Render(entry)
+		}
+	}
+}
```
rminnich, line 441: I hate to ask but ... is there a test?

GBA-17: Good idea, I can add a few tests.

pkg/menu/menu.go
GBA-17: Sounds good, I'll probably use the iterator version just for clarity.

rminnich: ok but def write a test for it then to cover 0 length, > 0 with no newline, only newline, etc.

rminnich (approved):
nice!

PR #177, opened 2020-09-24
Check for file before attempting to download

GBA-17:
We are not currently checking if an ISO already exists before we download it. This PR adds the logic to check for this.

Suppose the user requests to download an ISO, but we already have a copy. There are 3 cases to consider:

- Checksum file was not found
- Checksum file exists, but does not match ISO
- Checksum matches the ISO

Based on the case, display a message to the user and ask if they would like to continue with the download.

- If yes, erase the current ISO and its checksum file (if exists) and redownload
- If no, return to the main menu

### DIFF
--- a/cmds/webboot/utils.go
+++ b/cmds/webboot/utils.go
@@ -135,3 +135,20 @@ func validURL(url string) (string, string, bool) {
 		return url, "Invalid URL.", false
 	}
 }
+
+// TODO: Verify that this logic is valid
+func fileExists(filepath string) bool {
+	info, err := os.Stat(filepath)
+	return err == nil && info.Mode().IsRegular()
+}
+
+func checksumInfo(isoPath string) (string, string) {
+	checksumTypes := []string{"md5", "sha256"}
+	for _, checksumType := range checksumTypes {
+		checksumFile := isoPath + "." + checksumType
+		if fileExists(checksumFile) {
+			return checksumFile, checksumType
+		}
+	}
+	return "", ""
+}
--- a/cmds/webboot/webboot.go
+++ b/cmds/webboot/webboot.go
@@ -158,6 +158,15 @@ func (d *DownloadOption) exec(uiEvents <-chan ui.Event, network bool, cacheDir s
 		fpath = filepath.Join(downloadDir, filename)
 	}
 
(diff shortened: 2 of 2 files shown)


cmds/webboot/utils.go
```diff
@@ -114,3 +114,23 @@ func validURL(url string) (string, string, bool) {
 		return url, "Invalid URL.", false
 	}
 }
+
+func fileExists(filepath string) bool {
+	info, err := os.Stat(filepath)
+	if err != nil {
+		return false
+	} else {
+		return !info.IsDir()
```
rminnich, line 123: This is dangerous. I think you should do a full check for it being a regular file.

Also, when you have a return as on line 121, the convention is that you don't use an else.

since it is a bool you might even do
return err != nil && info.IsRegular()
on line 120

GBA-17: That makes sense, I updated the function to this:

```go
func fileExists(filepath string) bool {
	info, err := os.Stat(filepath)
	return err == nil && info.Mode().IsRegular()
}
```

cmds/webboot/utils.go
```diff
@@ -114,3 +114,19 @@ func validURL(url string) (string, string, bool) {
 		return url, "Invalid URL.", false
 	}
 }
+
+func fileExists(filepath string) bool {
```
rminnich, line 118: I'm wrong again .... I'm sorry.

If it exists at all I guess we do want to remove it -- if it's a dir, symlink, whatever, it has to be removed and we have to ask them. 

This is  harder than I realized.

GBA-17:
I think the logic in `checkDownloadRequired()` ended up too dense, but the idea was that there's 3 cases to check for.

1. The ISO exists, but there is no checksum file
2. The ISO and checksum file exist, but the checksum does not match
3. The ISO and checksum file exist, and the checksum matches

In all 3 cases, we give the user 2 options.

1. Erase the existing files and redownload
2. Keep the existing files and go back to the main menu

GBA-17:
I think the other issue with this PR is dealing with the file system. There's some complexities with checking if a file exists and removing a file that I'm not sure I handled correctly.

PR #179, opened 2020-09-25
Download checksum file along with ISO

GBA-17:
Download the checksum file when we download the ISO. This involved the following changes:

- Add a `checksumUrl` and `checksumType` field to each entry in the `supportedDistros` table
- Add the `DownloadInfo` struct to store information when we do multiple downloads
- In the `DownloadOption` menu flow, be sure to collect the link to the checksum file and download that file

Note that the naming convention for the checksum files is isoName.checksumType. For example, if we have `ubuntu.iso`, and it uses a sha256 checksum, the checksum file will be named `ubuntu.iso.sha256`

### DIFF
--- a/cmds/webboot/types.go
+++ b/cmds/webboot/types.go
@@ -15,6 +15,8 @@ type Distro struct {
 	isoPattern    string
 	bootConfig    string
 	kernelParams  string
+	checksumUrl   string
+	checksumType  string
 	customConfigs []bootiso.Config
 }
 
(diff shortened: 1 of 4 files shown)


GBA-17:
I'm wondering if there's a better way to ask for the checksum in the Custom Distro case. Right now if the user wants to download their own ISO, they need to type out the ISO url, Checksum url, and select the checksum type.

PR #182, opened 2020-09-29
Added LoadCustomConfigs() to the bootiso package

GBA-17:
We currently do not support ISOs with difficult-to-parse config files. One way around this is to manually specify the following information, which is usually parsed from the config file:

- Path to kernel
- Path to initrd
- Kernel parameters

This PR adds the `Config` type to the bootiso package, where we can specify these details.

In addition, the `LoadCustomConfigs()` function was added. This essentially does the job of `ParseConfigFromISO()`, except the boot configurations are specified in a list of `Config` objects rather than a config file.

Finally, a new test was added that runs `LoadCustomConfigs()` on our TinyCore ISO and a list of `Config` objects.

### DIFF
--- a/pkg/bootiso/bootiso.go
+++ b/pkg/bootiso/bootiso.go
@@ -10,6 +10,7 @@ import (
 	"hash"
 	"io"
 	"io/ioutil"
+	"log"
 	"os"
 	"path"
 	"strings"
@@ -23,10 +24,17 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// ParseConfigFromISO mounts an iso file to a
-// temp dir to get the config options
+type Config struct {
+	Label      string
+	KernelPath string
+	InitrdPath string
+	Cmdline    string
+}
+
+// ParseConfigFromISO mounts the iso file, attempts to parse the config file,
+// and returns a list of bootable boot.OSImage objects representing the parsed configs
 func ParseConfigFromISO(isoPath string, configType string) ([]boot.OSImage, error) {
-	tmp, err := ioutil.TempDir("", "mnt")
+	tmp, err := ioutil.TempDir("", "mnt-")
 	if err != nil {
 		return nil, fmt.Errorf("Error creating mount dir: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/bootiso/bootiso.go
```diff
@@ -43,12 +51,56 @@ func ParseConfigFromISO(isoPath string, configType string) ([]boot.OSImage, erro
...
+
+	mp, err := loopdev.Mount(tmpDir, unix.MS_RDONLY|unix.MS_NOATIME)
+	if err != nil {
+		return nil, err
+	}
+	defer mp.Unmount(0)
+
+	var images []boot.OSImage
+	for _, c := range configs {
+		kernelReader := uio.NewLazyFile("file://" + tmpDir + c.KernelPath)
```
rminnich, line 85: you should be using path.Join maybe with URL? Not filepath.Join, but path.Join
package main

import "fmt"
import "path"

func main() {
	fmt.Println(path.Join("file://", "x", "y"))

}

but there's a deeper problem here. LazyFiles do not get opened until the first read. 

So the file is not open when you leave this function I believe. 

The mount will unmount the iso.

the open on first read should then fail.

Also, for safety, your deferred functions maybe ought to have some error checking.

GBA-17: I was just noticing the problem with the Lazy Files. What would be the better way to open a reader on the kernel and initrd files?

GBA-17:
I copied the kernel and initrd's to tmp, but they need to stay open for kexec.

Are we allowed to call `defer os.Remove()` on the temp files, or is it ok to just keep them around?

pkg/bootiso/bootiso.go
```diff
@@ -43,12 +51,102 @@ func ParseConfigFromISO(isoPath string, configType string) ([]boot.OSImage, erro
...
+	if err != nil {
+		return nil, err
+	}
+
+	var images []boot.OSImage
+	var files []*os.File
+
+	defer func() {
+		for _, f := range files {
+			if err = f.Close(); err != nil {
```
rminnich, line 86: I'd do a log.Print as you still want to try to do the unmount

rminnich (approved):
tricky stuff, this one seems about as good as we can do ...

rminnich (approved):

PR #187, opened 2020-10-01
Add support for Debian

GBA-17:
Add an entry to the `supportedDistros` table for Debian. Also update the Supported Distros section of the README.


GBA-17:
This should pass the test if we rerun with [PR 188](https://github.com/u-root/webboot/pull/188). The problem is that the list of distro bookmarks is longer than the number of lines in the numbered menu, and you can only choose an option from the menu if it is currently displayed on screen.

rminnich (commented):
This is pretty fantastic. Our tests caught a UI problem? 

That is the dream.

rminnich (approved):

PR #190, opened 2020-10-02
Updated README description and install sections

GBA-17:
This makes the following updates to the README:

- Add Arch and Manjaro to the supported distros
- Describe the new custom config option
- Add a step for creating the ISO cache on the USB stick


orangecms (approved):

orangecms:
Really nice work! :) I hope my previous research on distro ISO structure helped a bit. Let me know if I can help out somewhere.
We can add more parsers to u-root, for example. :)

PR #191, opened 2020-11-28
Tweak pmem in syslinux.cfg.example to start at 512M

rminnich:
Some chromebooks don't have a full memory region at 1G for 1G.
Pmem is so fragile that it locks up the kernel in that case.

Until we can stop using pmem, at least make it lock things up
less.

### DIFF
--- a/syslinux.cfg.example
+++ b/syslinux.cfg.example
@@ -6,6 +6,6 @@ TIMEOUT 1
 LABEL webboot
   KERNEL /boot/webboot
   INITRD /boot/webboot.cpio.gz
-  APPEND earlyprintk=tty0 earlyprintk=ttyS0,115200,keep console=ttyS0 console=tty0 memmap=1G!1G vga=ask
+  APPEND earlyprintk=tty0 earlyprintk=ttyS0,115200,keep console=ttyS0 console=tty0 memmap=1G!512M vga=ask
 
 


orangecms:
That would insufficient for many distros though, wouldn't it? The bootable USB images I checked, excluding Tinycore which doesn't support pmem out of the box anyway, are either at something close to 1G or far beyond. Can't we safely start at some higher offset instead, such as 4G? On the other hand, webboot users can just create their own syslinux config anyway, so I don't mind too much what we offer upstream. Instead, we should probably just document what the weird `"1G!1G"` notation means and reference upstream docs (should they exist).

bjakobson:
@orangecms good point. 4G should work as well.

rminnich:
What kind of laptops are you people using :-)
this notation says 1G memory starting at 512m. This will work on the 5 systems I am testing on. 1G at 1G fails badly on my one chromebook that has memory in the range not quite to 2G. 

but I feel this is just fine. It takes for-f'ing-ever to download even 1G, and most of the distros that need that big a DVD don't work all that well with pmem. 

but 4G? starting at 4G? You have that much RAM? and 4G size? you just ruled out many chromebooks. 

I thinks this 1G at 512M is a fine compromise.

orangecms:
Damn, so I got the order wrong here...? I thought it meant 512M starting at 1G. Stupid notation. -.-

That is fine, I think. :-)

Regarding laptops: Yeah I am sorta taking 8G for granted these days, giben that I have 32G in my older Kaby Lake. 😅

orangecms (approved):

PR #195, opened 2021-05-23
distros: add MicroOS

orangecms:
Signed-off-by: Daniel Maslowski <info@orangecms.org>


orangecms:
I can't get it to work yet; this should work, but doesn't, from what I tried:
https://github.com/cyberorg/live-fat-stick/blob/master/live-grub-stick#L296

rjoleary:
> I can't get it to work yet; this should work, but doesn't, from what I tried:
> https://github.com/cyberorg/live-fat-stick/blob/master/live-grub-stick#L296

What goes wrong? Is it just an empty serial console?

orangecms:
The first initramfs from MicroOS doesn't find the ISO. I've tried other versions and also other openSUSE ISOs to no avail.

PR #196, opened 2021-05-23
add log option to the menu

orangecms:
- distros: add MicroOS
- oldwebboot: make it compile again
- Output verbose information on attempt to kexec


cmds/webboot/webboot.go
```diff
@@ -50,6 +51,10 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
 		distro = supportedDistros[entry.Label()]
 	}
 
+	verbose("Using distro %s with boot config %s", distroName, distro.bootConfig)
+
+	// Magic trick: The parsers may write very long and verbose output; too much!
+	log.SetOutput(&nilBuffer)
```
rminnich, line 57: I don't like just throwing  output away. A file in /tmp, fine. But just throwing it away makes debugging really hard.

cmds/webboot/webboot.go
```diff
@@ -267,6 +278,11 @@ func handleError(err error) {
 	wifiStderr.Reset()
 }
 
+func showLog() {
+	menu.DisplayResult(strings.Split(logBuffer.String(), "\n"), ui.PollEvents())
```
rminnich, line 282: but you threw lots of output away above. I assume this is a WIP?

orangecms: Sort of. The extensive output causes it to get lost entirely, and from within the interactive menu, there is zero feedback otherwise, which is why I started this here in the first place. Yes, it's a WIP, and I think it would make sense to offer another CLI arg for a path to a log file. The menu is not suitable for long text. Especially because scrolling down leaves the scroll output, so it's lost.

orangecms: Or in other words: Feedback within the interactive menu should be short enough to navigate it, and extensive output (the "full" version) written elsewhere, like a file.

orangecms: Okay I have a simple idea now, thank you!

cmds/webboot/webboot.go
orangecms: I'll get back to the file thing. For now I'll keep this as is.

cmds/webboot/webboot.go
```diff
@@ -106,10 +109,14 @@ func (i *ISO) exec(uiEvents <-chan ui.Event, boot bool) error {
 			s := fmt.Sprintf("config.image %s, kernelparams.String() %s", config.image, kernelParams.String())
 			return fmt.Errorf("Booting is disabled (see --dryrun flag), but otherwise would be [%s].", s)
 		}
-		err = bootiso.BootCachedISO(config.image, kernelParams.String() + " waitusb=10")
+		err = bootiso.BootCachedISO(config.image, kernelParams.String()+" waitusb=10")
 	}
 
 	// If kexec succeeds, we should not arrive here
+	if err == nil {
+		err = fmt.Errorf("kexec failed, but gave no error. Be sure to use kexec-tools.")
```
rminnich, line 117: This won't help most people, is there some way to make it a bit more informative? This is a tough case to explain to users.

rminnich (approved):
April was very careful to write tests for the UI, if you can think of some way to test this it would be nice. I can't.

cmds/webboot/webboot.go
orangecms: Definitely, you are raising the most important point. And this statement is even ~~misleading~~ wrong, since the scenario I tested _was_ actually using kexec-tools, but via cpu. If we cannot tell _what_, I would at least want to tell _where_ the issue was, since the boot mechanism involves so many steps.

orangecms:
> April was very careful to write tests for the UI, if you can think of some way to test this it would be nice. I can't.

I will check the test suite. My basic approach is to stub a kexec binary and let it just return, then check if webboot printed feedback by spying on the menu tool. It's an integration test scenario.

cmds/webboot/webboot.go
orangecms: I've rephrased it to be less suggestive about whether kexec-tools was used and left a TODO.

rjoleary (approved):

PR #219, opened 2021-06-08
update error messages

kellysun878:
Update error messages to show stack trace


rminnich (approved):

rminnich (dismissed):
You'll need to fix the commit message. 

Right now it is
...e errro

in 2 years, what will this tell someone? A good thing is to show sample output, maybe. 
Also the subject should not end in ..., maybe it could be

update error messages to show stack trace

orangecms (approved):

PR #225, opened 2021-06-10
Add travis CI check for `go fmt`

rjoleary:
Signed-off-by: Ryan O'Leary <ryanoleary@google.com>

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -e
 
+# Check that the code has been formatted correctly.
+test -z "$(gofmt -s -l *.go pkg cmds)"
+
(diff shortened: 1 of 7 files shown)


travis.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -e
 
+# Check that the code has been formatted correctly.
+test -z "$(gofmt -l $(find -name '*.go' | grep -v /vendor/))"
```
rminnich, line 5: you can do this but it's doubtless slower than simply
gofmt -s -l *.go pkg cmds
?
I just tried that and it works fine.

rminnich (approved):

PR #228, opened 2021-06-16
Automatically download and compile wpa_supplicant

rjoleary:
To download and build your own wpa_supplicant, use:

	go run . --wpa-version=2.9

For me, this works for 2.7, 2.8 and 2.9.

Signed-off-by: Ryan O'Leary <ryanoleary@google.com>


rminnich:
What problem does this solve?

rminnich:
While I realize this builds, you're just substituting one unknown for another, so I'm still not sure of the value. Does this "build from source" come with a corresponding increase in understanding of what is going on?

rjoleary:
Let's solve the other problem first.

This solves a different problem from what you're probably thinking. It solves the problem of different people are using different versions of wpa_supplicant.

rminnich (approved):

rjoleary:
> Let's solve the other problem first.
> 
> This solves a different problem from what you're probably thinking. It solves the problem of different people are using different versions of wpa_supplicant.

This change is basically a no-op unless you set `--wpa-supplicant=2.7`, so it is safe to submit.

rjoleary:
For some reason github actions is not happy: `Error: fatal: unable to access 'https://github.com/u-root/webboot/': The requested URL returned error: 500`

rjoleary:
> For some reason github actions is not happy: `Error: fatal: unable to access 'https://github.com/u-root/webboot/': The requested URL returned error: 500`

GitHub is happy again. :)

rminnich (dismissed):

rminnich (approved):

Issue #247, opened 2021-07-01
Ubuntu: no space left on device


rjoleary:
**Describe the bug**
The ISO download is interrupted with the error "write /tmp/iso-download-758024287: no space left on device".

**To Reproduce**
On an x230 laptop, select the Ubuntu distro and wait while it downloads. The laptop has 4GB of RAM.

rjoleary:
My USB driver is one large partition:

```
Device     Boot Start       End   Sectors  Size Id Type
/dev/sda1  *     2048 121061375 121059328 57.7G ef EFI (FAT-12/16/32)
```

This is more than enough to hold the image. I have a feeling we are downloading to RAM instead of the USB stick.

kellysun878:
Fixed by #272.

PR #272, opened 2021-07-29
Download ISOs to USB instead of RAM

kellysun878:
Previously, ISOs were downloaded to RAM then copied to the USB,
but this doesn't work for large ISOs because RAM runs out of space. 
Now, ISOs are downloaded directly to the USB.


rjoleary (approved):

rjoleary:
Fixes #247

rjoleary (approved):

PR #283, opened 2021-08-10
Remove one of distroData()'s return values

kellysun878:
Previously, distroData() returned a map[string]Distro, but this is not
necessary because supportedDistros is a map[string]Distro and is a
global variable.


rjoleary (changes requested):
Try to do this without a global variable.

PR #304, opened 2022-06-20
Fix modules and remove NiChrome dependency

paulzzy:
- Delete vendors
- Modules: Add `go.mod`, `go.sum`
- Remove dependency on `github.com/u-root/NiChrome`, use files in this repo


pkg/menu/menu_test.go
```diff
@@ -367,16 +367,6 @@ func TestDisplayMenu(t *testing.T) {
 				pressKey(uiEvents, []string{"2", "a", "<Backspace>", "<Enter>"})
 			},
 		},
-		{
```
rminnich, line 370: I'd prefer that we comment this out, rather than remove it, because I'd like this test to work some day.

pkg/menu/menu_test.go
```diff
@@ -367,16 +367,16 @@ func TestDisplayMenu(t *testing.T) {
...
-			name:    "<pageDown>_<pageUp>_<pageDown>_hit_11",
-			entries: []Entry{entry1, entry2, entry3, entry4, entry5, entry6, entry7, entry8, entry9, entry10, entry11, entry12},
-			// hit <pageDown> -> <pageUp> -> <pageDown> current page is : 0~9
-			want: entry12,
-			human: func(uiEvents chan ui.Event, menus <-chan string) {
-				nextMenuReady(menus)
-				pressKey(uiEvents, []string{"<PageDown>", "<pageUp>", "<PageDown>", "1", "1", "<Enter>"})
-			},
-		},
+		// {
```
rminnich, line 370: OH I FORGOT :-(
put a link here to the bug you found in termui, along with a comment about why this is commented out.
thanks

rminnich (approved):

PR #309, opened 2022-07-29
Fix nightly.yml

paulzzy:
- Do not build the kernel every night; just wget it
- The vmtest package is a non-working thing; bypass most of it
- Since u-root is all about paths, to a shallow clone into this directory and use it
- build the u-root we git clone
- Build a u-root and use it
- Finish up fixing nightly
- Fix paths in nightly.yml

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -31,11 +31,11 @@ jobs:
       - name: Kernel Build
         run: |
           ./nightly.sh
-          ls -l linux/arch/x86/boot/bzImage
+          ls -l bzImage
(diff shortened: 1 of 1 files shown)


rminnich (commented):
can you see if we need the .DS_STORE file (I can't see why?) and the logOutput.txt file?

rminnich (changes requested):
yeah def remove those 2 files I think?

paulzzy:
My bad, I forgot to check what I was adding.

paulzzy:
Ideally those would be in .gitignore, but that can be a separate PR.

rminnich (approved):

PR #311, opened 2022-08-26
Fix integration testing, add to CI

paulzzy:
- Run tests with standalone integration.sh
- Integration testing workflow now in ci.yml instead of nightly.yml


ryukinix (approved):

paulzzy:
I updated all distro links, so downloading ISOs shouldn't be a problem. However, most distros (except TinyCore) have outdated boot info that causes the tests to fail. I'll leave that as a TODO, since the integration testing logic is now fixed and only config info is wrong.

rminnich:
what should we do about the failures? merge anyway?

rminnich:
oh I read your comment. Suggestion: add a map called broken, and add these to it, and VERY LOUDLY log that they are being skipped? If you're out of time, I'll do that.

paulzzy:
That's a good suggestion! I am out of time, so I would greatly appreciate if you could implement it.

rminnich (approved):

